PR #24, opened 2021-11-08
Feat/finish new project

oskarikerppo:


components/projects/NewProject/NewProjectForm.tsx
```diff
@@ -29,9 +30,32 @@ function NewProjectForm({ employees, customers }: NewProjectFormProps): JSX.Elem
         }
     };
 
-    const handleSumbit = (e: React.FormEvent<HTMLFormElement>) => {
+    const handleSubmit = async (e: React.FormEvent<HTMLFormElement>) => {
         e.preventDefault();
-        console.log(state);
+        // console.log(state);
```
LeadingMoominExpert, line 35: Remove all // console.logs, they serve no purpose on the main branch

components/projects/NewProject/NewProjectForm.tsx
```diff
@@ -29,9 +30,32 @@ function NewProjectForm({ employees, customers }: NewProjectFormProps): JSX.Elem
...
+        };
+        const url = `${process.env.NEXT_PUBLIC_API_URL}/project`;
+        const res = await fetch(url, {
+            method: 'POST',
+            headers: { 'Content-Type': 'application/json' },
+            body: JSON.stringify(data),
+        });
+        const placehodler = await res.json();
+        // console.log(res.status);
+        // console.log(placehodler);
```
LeadingMoominExpert, line 58: same

components/projects/NewProject/NewProjectForm.tsx
```diff
@@ -29,9 +30,32 @@ function NewProjectForm({ employees, customers }: NewProjectFormProps): JSX.Elem
...
+                first_name: state.managingEmployee?.first_name,
+                last_name: state.managingEmployee?.last_name,
+                email: state.managingEmployee?.email,
+            },
+        };
+        const url = `${process.env.NEXT_PUBLIC_API_URL}/project`;
+        const res = await fetch(url, {
+            method: 'POST',
+            headers: { 'Content-Type': 'application/json' },
+            body: JSON.stringify(data),
```
LeadingMoominExpert, line 54: Can't you use just the `data` as is.

components/projects/NewProject/NewProjectForm.tsx
```diff
@@ -29,9 +30,32 @@ function NewProjectForm({ employees, customers }: NewProjectFormProps): JSX.Elem
         }
     };
 
-    const handleSumbit = (e: React.FormEvent<HTMLFormElement>) => {
+    const handleSubmit = async (e: React.FormEvent<HTMLFormElement>) => {
         e.preventDefault();
-        console.log(state);
+        // console.log(state);
+        const data = {
```
LeadingMoominExpert, line 36: Can you infer the type of `data` ? Also the variable could have a better name. `createProjectRequest` or something that gives a hint what it's used for
```typescript
const createProjectRequest: ProjectDTO = {
....
}
```

components/projects/NewProject/NewProjectForm.tsx
```diff
@@ -29,9 +30,32 @@ function NewProjectForm({ employees, customers }: NewProjectFormProps): JSX.Elem
...
+            endData: state.endDate,
+            customer: { id: state.customer?.id, name: state.customer?.name },
+            managingEmployee: {
+                id: state.managingEmployee?.id,
+                first_name: state.managingEmployee?.first_name,
+                last_name: state.managingEmployee?.last_name,
+                email: state.managingEmployee?.email,
+            },
+        };
+        const url = `${process.env.NEXT_PUBLIC_API_URL}/project`;
```
LeadingMoominExpert, line 50: `${process.env.NEXT_PUBLIC_API_URL}` should be a high level private const definition. Having it here in the middle of the method makes the code hard to read

lib/hooks/useCustomers.ts
```diff
@@ -10,9 +10,9 @@ interface CustomerResponse {
 
 function useCustomers(): CustomerResponse {
     // mock endpoint
-    const endpoint = 'http://localhost:3000/api/customers';
```
LeadingMoominExpert, line 13: Variables that aren't needed should be deleted, not commented out and left in commits.

lib/hooks/useEmployees.ts
```diff
@@ -10,9 +10,9 @@ interface EmployeeResponse {
 
 function useEmployees(): EmployeeResponse {
     // actual endpoint
-    // const endpoint = process.env.NEXT_PUBLIC_API_URL + '/employee';
+    const endpoint = process.env.NEXT_PUBLIC_API_URL + '/employee';
     // mock endpoint
-    const endpoint = 'http://localhost:3000/api/employees';
+    // const endpoint = 'http://localhost:3000/api/employees';
```
LeadingMoominExpert, line 15: Same, can be removed

lib/hooks/useCustomers.ts
```diff
@@ -10,9 +10,9 @@ interface CustomerResponse {
...
+    // const endpoint = 'http://localhost:3000/api/customers';
     // actual endpoint
-    // const endpoint = process.env.NEXT_PUBLIC_API_URL + '/customer';
+    const endpoint = process.env.NEXT_PUBLIC_API_URL + '/customer';
```
LeadingMoominExpert, line 15: I'd use either concatenation or string templating throughout the project, not both here and there. Like in the NewProjectForm this could be 
```typescript
const endpoint = `${process.env.NEXT_PUBLIC_API_URL}/customer`;
```

LeadingMoominExpert (changes requested):
Frontend pull requests should have screenshots of what has been changed

components/projects/NewProject/NewProjectForm.tsx
robertsiipola: If you mean omitting the `JSON.stringify` function then no. The backend will throw the following error:

`JSON parse error: Cannot deserialize value of type `three.consulting.epoc.dto.ProjectDTO` from Array value (token `JsonToken.START_ARRAY`); nested exception is com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot deserialize value of type `three.consulting.epoc.dto.ProjectDTO` from Array value (token `JsonToken.START_ARRAY`)<LF> at [Source: (PushbackInputStream); line: 1, column: 1]]`

components/projects/NewProject/NewProjectForm.tsx
robertsiipola: Yes. And in case there's a problem with using all the fields (probably won't be), you can always use a partial type:

`const createProjectRequest: Partial<components['schemas']['ProjectDTO']> = { ... }`

components/projects/NewProject/NewProjectForm.tsx
robertsiipola: We also want some kind of response to both a successful and an unsuccessful post request. A response object has a key called `status` indicating its... well, status. For a successful `POST` request, this value will be `201` (see MDN documentation for [status codes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status) and for the fetch [response object](https://developer.mozilla.org/en-US/docs/Web/API/Response))

One can then write an if/else clause after `res` that checks the status code. If it's `201` we should change the state of `FormStatus` to be `SUCCESSFUL` and direct the user to the `/projects` page. You can do this by using the `router` object in Next, and its push function (see [here](https://nextjs.org/docs/api-reference/next/router))

If the server responds with anything else than a `201`, set the `FormStatus` field to `ERROR`. Additionally, add an error indicator to the JSX-code that only renders if the `FormStatus` is `ERROR`. It should tell the user that submitting the form failed.

robertsiipola (changes requested):

components/projects/NewProject/NewProjectForm.tsx
oskarikerppo: Fixed in df7dea1

components/projects/NewProject/NewProjectForm.tsx
oskarikerppo: Fixed in df7dea1

lib/hooks/useEmployees.ts
oskarikerppo: Fixed in 397993e

lib/hooks/useCustomers.ts
oskarikerppo: Fixed in 397993e

components/projects/NewProject/NewProjectForm.tsx
```diff
@@ -3,6 +3,7 @@ import reducer, { init, initialState, ActionType, FormStatus } from './reducer';
 import { Flex } from '@chakra-ui/layout';
 import React, { useReducer } from 'react';
 import { components } from '@/lib/types/api';
+import { State } from 'swr/dist/types';
```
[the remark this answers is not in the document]
oskarikerppo: Fixed in 3f81e47

lib/hooks/useCustomers.ts
oskarikerppo: Fixed in bdfadda

components/projects/NewProject/NewProjectForm.tsx
oskarikerppo: Fixed in bdfadda

components/projects/NewProject/NewProjectForm.tsx
oskarikerppo: This is done in 5bf47e0. Succesful POST request sends user back to /projects (status 200 or 201). Any other status is handled by rendering a Box below Submit button displaying some text. What should happen in case of error should be elaborated.

components/projects/NewProject/NewProjectForm.tsx
oskarikerppo: Done in 5bf47e0

components/projects/NewProject/NewProjectForm.tsx
oskarikerppo: Screenshot from 2021-11-09 15-50-27

LeadingMoominExpert (approved):

robertsiipola (approved):

PR #26, opened 2021-11-10
Chore/gha

robertsiipola:
Adds simple Github Actions that runs the typescript compiler and prettier whenever commits are pushed and on pull requests.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,12 +4,14 @@ jobs:
     tsc-and-format:
         runs-on: ubuntu-latest
         steps:
-            - uses: actions/checkout@v1
-            - name: install node v12
-              uses: actions/setup-node@v1
+            - name: Checkout code
+              uses: actions/checkout@v2
+
+            - name: Setup node
+              uses: actions/setup-node@v2
               with:
-                  node-version: 12
-            - name: yarn install
-              run: yarn install
-            - name: tsc
-              uses: icrawl/action-tsc@v1
+                  node-version: '14'
+            - name: Install dependencies
+              run: |
+                  yarn install
+                  yarn run type-check
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,5 +1,15 @@
 name: housekeeping
-on: [push]
+on: [push, pull_request]
 jobs:
     tsc-and-format:
         runs-on: ubuntu-latest
+        steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -0,0 +1,18 @@
+name: housekeeping
+on: [push, pull_request]
```
LeadingMoominExpert, line 2: The triggers should be more excplicit, not sure what the defaults are on e.g. `pull_request`. You can do something like
```yaml
on:
  push:
    branches:
    - '**'
  pull_request:
    types:
      - opened
      - synchronize
      - reopened  
```
And in the future if you wan't to exclude `main` branch from this workflow and run some deployment or something else there, you can add `- '!main'` under branches.

.github/workflows/main.yml
```diff
@@ -0,0 +1,18 @@
...
+jobs:
+    tsc-and-format:
+        runs-on: ubuntu-latest
```
LeadingMoominExpert, line 5: Below `runs-on` you should add a custom timeout, because the gha default is one hour and if the build fails from a mysterious reason it won't fail quickly enough
```yaml
timeout-minutes: 10
```

LeadingMoominExpert (changes requested):

.github/workflows/main.yml
```diff
@@ -0,0 +1,18 @@
...
+        steps:
+            - name: Checkout code
+              uses: actions/checkout@v2
+
+            - name: Setup node
+              uses: actions/setup-node@v2
+              with:
+                  node-version: '14'
+            - name: Install dependencies
```
LeadingMoominExpert, line 14: This name could also be more descriptive as you also run two other commands after install

.github/workflows/main.yml
robertsiipola: Added this in 1cbf73c.

.github/workflows/main.yml
robertsiipola: Added this in 1cbf73c

.github/workflows/main.yml
robertsiipola: Changed it to `Install dependencies and run tsc and prettier` in 1cbf73c.

LeadingMoominExpert (approved):

oskarikerppo (approved):

PR #39, opened 2021-11-15
Feat/new customer modal

oskarikerppo:


oskarikerppo:
Screenshot from 2021-11-15 15-12-35

A little margin is added between Add Customer button and Select customer drop-down menu

LeadingMoominExpert (commented):
Please add screenshots of changes before review 

edit: ja sieltä

oskarikerppo:
Screenshot from 2021-11-15 15-15-27

Images of the modal window
)

.eslintignore
```diff
@@ -1,3 +1,5 @@
 lib/types/api.ts
 lib/utils/fetcher.ts
+pages/_document.tsx
+components/common/ClientOnlyPortal.ts
```
LeadingMoominExpert, line 4: Not a fault of this PR but the file is missing a newline at the end

robertsiipola:
The `ErrorAlert` component looks a little dumb in such a small place, but I'd still approve the PR because that's a problem with the `ErrorAlert` component and can be fixed in another PR.

robertsiipola (approved):

components/projects/NewProject/NewCustomer.tsx
```diff
@@ -9,28 +11,92 @@ import {
...
+    const { mutate } = useSWRConfig();
+
+    const handleSubmit = async (e: React.MouseEvent) => {
+        e.preventDefault();
+
+        const createCustomerRequest: components['schemas']['CustomerDTO'] = {
+            name: state.name,
+            description: state.description,
+        };
+        setState({ name: state.name, description: state.description, formStatus: FormStatus.LOADING });
```
LeadingMoominExpert, line 42: What's the purpose of the first two `name: state.name` and `description: state.description`? I don't think setting the state to the same value does really anything and can be omitted

components/projects/NewProject/NewCustomer.tsx
```diff
@@ -9,28 +11,92 @@ import {
...
+        try {
+            await fetch.post(url, createCustomerRequest);
+            mutate(url);
+            setState({ name: state.name, description: state.description, formStatus: FormStatus.SUCCESS });
+            onClose();
+        } catch (error) {
+            setState({ name: state.name, description: state.description, formStatus: FormStatus.ERROR });
+        }
```
LeadingMoominExpert, line 50: This error should probably be added to some notification? Or are we just catching errors without trying to find out what the error is?

LeadingMoominExpert: One way would be to have `errorMessage: string` or similar as a state parameter, which is displayed if it has text

components/projects/NewProject/NewCustomer.tsx
```diff
@@ -9,28 +11,92 @@ import {
...
+                                placeholder="Customer Name"
+                                onChange={(e) =>
+                                    setState({
+                                        name: e.target.value,
+                                        description: state.description,
+                                        formStatus: FormStatus.IDLE,
+                                    })
+                                }
+                            />
+                            {console.log(state)}
```
LeadingMoominExpert, line 74: Please remove console.logs.

components/projects/NewProject/NewCustomer.tsx
```diff
@@ -9,28 +11,92 @@ import {
...
+                                placeholder="Description"
+                                onChange={(e) =>
+                                    setState({
+                                        name: state.name,
+                                        description: e.target.value,
+                                        formStatus: FormStatus.IDLE,
+                                    })
+                                }
+                            />
+                            {console.log(state)}
```
LeadingMoominExpert, line 89: Same

components/projects/NewProject/NewProjectForm.tsx
```diff
@@ -121,18 +121,23 @@ function NewProjectForm({ employees, customers }: NewProjectFormProps): JSX.Elem
...
-                            );
-                        })}
-                    </Select>
-                </FormControl>
+                <Flex flexDirection="row" justifyContent="center">
+                    <FormControl isRequired={true}>
+                        <FormLabel>Customer</FormLabel>
+                        <Flex flexDirection="row" justifyContent="space-between">
+                            <Select onChange={handleCustomerChange} placeholder="Select customer" marginRight="0.3rem">
+                                {customers?.map((el, idx) => {
```
LeadingMoominExpert, line 129: I think you could use more descriptive variable naming, unless it's a preference to omit it? It's an array of customers so you could use 
```typescript
customers?.map((customer, idx) => {
```
etc

tsconfig.json
```diff
@@ -20,5 +20,5 @@
         }
     },
     "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx"],
-    "exclude": ["node_modules"]
+    "exclude": ["node_modules", "components/common/ClientOnlyPortal.ts"]
```
LeadingMoominExpert, line 23: What is this file and why is it excluded?

LeadingMoominExpert (changes requested):

components/projects/NewProject/NewCustomer.tsx
oskarikerppo: I think the state will only contain the formStatus value unless name and description are explicitly given

.eslintignore
oskarikerppo: Fixed in 9137767

components/projects/NewProject/NewCustomer.tsx
oskarikerppo: Added errorMessage to state in 9137767. Error message is shown in a Box element. I think it should be a part of ErrorAlert component, but that will have to wait for tomorrow

components/projects/NewProject/NewCustomer.tsx
oskarikerppo: removed

components/projects/NewProject/NewCustomer.tsx
oskarikerppo: removed

components/projects/NewProject/NewProjectForm.tsx
oskarikerppo: el changed to customer in 9137767

tsconfig.json
oskarikerppo: That file doesn't exist (anymore). It was never pushed to github, but I forgot to remove that exclusion. Fixed in 9137767

components/projects/NewProject/NewCustomer.tsx
LeadingMoominExpert: Strange behavior. `setState` should only affect the specified fields, maybe @robertsiipola know if there's something wrong with the state definition. 

Anyhoo, maybe you can use the spread operator with the existing state combined with setting the fields that are actually being change, e.g.
```typescript
setState({
    ...state,
    formStatus.LOADING
})
```
to only change the `formStatus` value?

LeadingMoominExpert: > setState should only affect the specified fields

I actually know jack shit about the hooks way of managing state. Apparently some even use own hooks for each state parameter. But spreading the existing state should be a good way to reduce some repetitive code

oskarikerppo: spread operator put to use in 296a601

LeadingMoominExpert (approved):

PR #42, opened 2021-11-18
Feat/view for single projects

oskarikerppo:


oskarikerppo:
I'm opening this pull request, although it still needs some work but I need some comments

oskarikerppo:
Screenshot from 2021-11-18 09-59-24

Edit window for project

oskarikerppo:
Screenshot from 2021-11-18 09-59-11

What information should be displayed on this windows? What type of elements should be used? How should the page look like?

robertsiipola:
> Screenshot from 2021-11-18 09-59-11
> 
> What information should be displayed on this windows? What type of elements should be used? How should the page look like?

For now the fields that would be necessary are `name`, `description`, `customer?.name`, `start_date`, `end_date`, `employee?.name` and `project.status`.

For the styling, use a similar "card" layout as with the forms (see the edit form you uploaded). Use a `Heading` component from Chakra that has the name of the project. Under that layout the rest of the fields, starting with the description of the project.

I'll put some thought into a proper design later, but you can get started with this now.

oskarikerppo:
Screenshot from 2021-11-18 13-36-28

Modified the layout of project details in ca5bf41.

I also noticed that the end date can precede the start date. This is probably not intended behavior

LeadingMoominExpert:
One thing that comes to mind is that clicking open a project to edit it opens a modal with empty values, and not the existing ones.

> I also noticed that the end date can precede the start date. This is probably not intended behavior

I'll create an issue on that

robertsiipola (changes requested):
Instead of edit showing an empty form, I'd want its state to be initialised by the project data we want to edit.

You could do this in the following way:

1. Give `NewProjectForm` an optional prop called `project`, which is a project object. `project` gets passed to the form only when we're editing an existing project object.
2. You can query the project to be edited in the page component (`/pages/{id}/edit`). Pass that project to the form.
3. Inside the form, in the `useReducer` hook you can create a conditional that checks if project exists and if it does it will use the project prop as the initial state. Check [this](https://gist.github.com/robertsiipola/fa012183de66dfbe7c3ce01e60720fc0) gist for an example.
4. Also change the name of the `ProjectForm` file to `ProjectForm.tsx` from `NewProjectForm.tsx`.

oskarikerppo:
> Instead of edit showing an empty form, I'd want its state to be initialised by the project data we want to edit.
> 
> You could do this in the following way:
> 
> 1. Give `NewProjectForm` an optional prop called `project`, which is a project object. `project` gets passed to the form only when we're editing an existing project object.
> 2. You can query the project to be edited in the page component (`/pages/{id}/edit`). Pass that project to the form.
> 3. Inside the form, in the `useReducer` hook you can create a conditional that checks if project exists and if it does it will use the project prop as the initial state. Check [this](https://gist.github.com/robertsiipola/fa012183de66dfbe7c3ce01e60720fc0) gist for an example.
> 4. Also change the name of the `ProjectForm` file to `ProjectForm.tsx` from `NewProjectForm.tsx`.

Done in e37d31c

LeadingMoominExpert (changes requested):
What does hitting `submit` when editing project do?
image
All the data is present already so why run 3x GET first? Especially when two of those are getting ALL entitites in database. IMO only the single submit should be ran. It'll slow down the app too much if we query everything beforehand.

Also I'd like to have a `cancel` or `go back` or a button like that next to the `submit`, which takes the user back to the all projects view

LeadingMoominExpert (approved):

robertsiipola (approved):

PR #52, opened 2021-11-25
refactor: generic useData hook for retrieving data from backend

oskarikerppo:


oskarikerppo:
In useData hook, I think the generic DataResponse type "data" and the const "ep" in function useData could be named better. Any suggestions?

lib/hooks/useData.tsx
```diff
@@ -0,0 +1,16 @@
...
+    isLoading: boolean;
+    isError?: Error;
+};
+
+function useData<T>(endpoint: string): DataResponse<T> {
+    const ep = `${process.env.NEXT_PUBLIC_API_URL}/${endpoint}`;
+    const { data: data, error } = useSWR<T, Error>(ep, fetch.get);
+    return { data, isLoading: !data && !error, isError: error };
+}
+
```
robertsiipola, line 15: Let's add a parameter called queryParams that get appended to the url. See this gist for an [example](https://gist.github.com/robertsiipola/ab9bb8d22cf5efb99194f3e9260c8aaa).

This way, we can later use methods like filters etc.

robertsiipola (changes requested):

lib/hooks/useData.tsx
oskarikerppo: added in 82846f0

pages/projects/[id]/index.tsx
```diff
@@ -10,11 +9,12 @@ import Layout from '@/components/common/Layout';
 import { Button } from '@chakra-ui/react';
 import Link from 'next/link';
 import { components } from '@/lib/types/api';
+import useData from '@/lib/hooks/useData';
 
 const Id: NextPage = () => {
     const router = useRouter();
     const { id } = router.query;
-    const { projects, isError, isLoading } = useProjects<components['schemas']['ProjectDTO']>(id);
+    const { data: projects, isError, isLoading } = useData<components['schemas']['ProjectDTO']>(`project/${id}`);
```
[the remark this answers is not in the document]
Comment (oskarikerppo) [reply pages/projects/[id]/index.tsx]: fixed in 82846f0

robertsiipola (approved):

PR #53, opened 2021-11-26
chore: added lint-staged

oskarikerppo:
I installed lint-staged via `npx mrm@2 lint-staged`

Then modified the settings a bit. I have no idea what I was doing so please check for errors.


robertsiipola:
I could not get the changes here to work for the life of my and I don't know why. I got it to work by installing lint staged with yarn and adding the configurations manually. I'll close this PR and opened another one in its stead. See #56.

PR #56, opened 2021-11-28
test: added lint-staged and the no-console rule for eslint

### LINKED ISSUE #30: Change linting to use the lint staged package.
The package runs lints (eslint, prettier) only on files that are staged, this makes the linting process faster.

### LINKED ISSUE #40: Add the `no-console` rule for eslint.
The rule can be found [here](https://eslint.org/docs/rules/no-console).

robertsiipola:
I could not get the changes in #53 to work for the life of my and I don't know why. I got it to work by installing lint staged with yarn and adding the configurations manually. Will close PR #53. Closes #30, closes #40.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -2,5 +2,4 @@
 . "$(dirname "$0")/_/husky.sh"
 
 yarn type-check
-yarn prettier --check .
-yarn eslint .
+yarn lint-staged
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -14,7 +14,7 @@ module.exports = {
         sourceType: 'module',
     },
     plugins: ['react', '@typescript-eslint'],
-    rules: {},
+    rules: { 'no-console': 'error' },
     settings: {
         react: {
             version: 'detect',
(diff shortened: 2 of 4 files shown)


oskarikerppo (approved):

.husky/pre-commit
```diff
@@ -2,5 +2,4 @@
 . "$(dirname "$0")/_/husky.sh"
 
 yarn type-check
-yarn prettier --check .
-yarn eslint .
\ No newline at end of file
+yarn lint-staged
```
LeadingMoominExpert, line 5: Missing a newline at the end of file. There seems to be a quite a bit of these in this repo, so I'd suggest creating a file called `.editorconfig` to the root of the repo with just the content
```
root = true

[*]
insert_final_newline=true
```
So you can automatically have \n at the end of each file.

LeadingMoominExpert (approved):

.husky/pre-commit
robertsiipola: Will open a new PR for this. I'm assuming this is mostly a problem with files that aren't being formatted by prettier (dot files and json files).

PR #58, opened 2021-12-01
Feat/linking users with timesheets

oskarikerppo:


oskarikerppo:
Screenshot from 2021-12-01 16-48-21

components/timesheets/TimesheetTable.tsx
```diff
@@ -0,0 +1,232 @@
...
+                <Button colorScheme="blue" onClick={onOpen}>
+                    Add User
+                </Button>
+            </Flex>
+            <Modal isOpen={isOpen} onClose={onClose}>
+                <ModalOverlay />
+                <ModalContent>
+                    <ModalHeader>Add user to project</ModalHeader>
+                    <ModalCloseButton />
+                    <FormControl>
```
robertsiipola, line 167: Please add left and right padding to this form. 0.5 rem should be fine on both sides.

robertsiipola (changes requested):

LeadingMoominExpert:
> The button "x" in the table doesn't do anything yet. TimesheetDTO needs a field "status" that can be set to "archived" first. Also there needs to be some check either in the backend or frontend that a user is not added multiple times, a users total allocation is not over 100, etc.

Backend validation for correct allocation already exists: https://github.com/three-consulting/epoc-backend/blob/main/src/main/kotlin/three/consulting/epoc/dto/TimesheetDTO.kt#L13

For the combination of employee and project to be unique on timesheet, we can do a multi parameter constraint like https://stackoverflow.com/a/2772510/7990795

components/timesheets/TimesheetTable.tsx
oskarikerppo: Screenshot from 2021-12-08 10-44-52
added in 727f11d

robertsiipola (approved):

PR #59, opened 2021-12-08
feat: added authorization header to all requests

oskarikerppo:


lib/utils/fetch.ts
```diff
@@ -1,6 +1,9 @@
 // https://eckertalex.dev/blog/typescript-fetch-wrapper
-async function http<T>(path: string, config?: RequestInit): Promise<T> {
-    const request = new Request(path, { headers: { 'Content-Type': 'application/json' }, ...config });
+async function http<T>(path: string, config?: RequestInit, auth?: string): Promise<T> {
+    const request = new Request(path, {
+        headers: { 'Content-Type': 'application/json', Authorization: `Bearer ${auth}` },
+        ...config,
+    });
```
LeadingMoominExpert, line 6: Can't you just init the `const idJwt = useUser()?.getSignInUserSession()?.getIdToken().getJwtToken();` variable here and set the Authorization header if it's not null? Then you wouldn't need to get the token on every form and pass it along to this function

LeadingMoominExpert (changes requested):

lib/utils/fetch.ts
oskarikerppo: `import { useUser } from '@/lib/hooks/useAuth';

// https://eckertalex.dev/blog/typescript-fetch-wrapper
async function http<T>(path: string, config?: RequestInit): Promise<T> {
    const idJwt = useUser()?.getSignInUserSession()?.getIdToken().getJwtToken();
    const request = new Request(path, {
        headers: { 'Content-Type': 'application/json', Authorization: `Bearer ${idJwt}` },
        ...config,
    });
    const response = await fetch(request);`

Doesn't seem to work

LeadingMoominExpert: How about 
```typescript
...
const authSession = await Auth.currentSession();
const jwt = authSession.getIdToken()?.getJwtToken();
```

oskarikerppo: That works

components/projects/NewProject/NewCustomer.tsx
```diff
@@ -37,6 +38,7 @@ function NewCustomer(): JSX.Element {
     });
     const url = `${process.env.NEXT_PUBLIC_API_URL}/customer`;
     const { mutate } = useSWRConfig();
+    const idJwt = useUser()?.getSignInUserSession()?.getIdToken().getJwtToken();
```
LeadingMoominExpert, line 41: Now you can remove the token definitions and useUser imports from the components

LeadingMoominExpert (changes requested):

LeadingMoominExpert (approved):

PR #63, opened 2022-01-05
fix: fixed error related to navbar

oskarikerppo:


oskarikerppo:
Screenshot from 2022-01-05 15-16-39

PR #65, opened 2022-01-11
feat: task table added to projects

oskarikerppo:


oskarikerppo:
Screenshot from 2022-01-11 15-24-16
Screenshot from 2022-01-11 15-24-38

oskarikerppo:
Screenshot from 2022-01-11 15-25-46

The modal warns the user if end date precedes start date. However, this will still go through as backend doesn't check for this.

oskarikerppo:
Currently only the task name is shown. How should the other info be diplayed? Seems like too much to open up another modal. Should clicking a task route to a new page where all info of tasks are shown?

oskarikerppo:
Also the "x" button doesn't do anything yet, as there is no "status" field in TaskDTO

components/tasks/TaskTable.tsx
```diff
@@ -0,0 +1,222 @@
...
+                                setState({
+                                    ...state,
+                                    startDate: e.target.value,
+                                })
+                            }
+                        ></Input>
+                        <FormErrorMessage>Task start date must not be empty.</FormErrorMessage>
+                    </FormControl>
+                    <FormControl
+                        isInvalid={state.startDate ? (state.endDate ? state.endDate <= state.startDate : false) : false}
```
LeadingMoominExpert, line 190: Is this correct that both ternary elses give false or am I just reading this incorrectly? If there's any way to avoid using nested ternary expressions I would avoid them because they are very hard to read

LeadingMoominExpert (changes requested):

components/tasks/TaskTable.tsx
oskarikerppo: It returns true if both dates are defined and end date precedes start date. Otherwise returns false. I don't know how to do it otherwise, as you must check that both are not undefined

oskarikerppo: Which I did in 66133e8. Is that better?

components/tasks/TaskTable.tsx
```diff
@@ -0,0 +1,222 @@
...
+                        <Input
+                            placeholder="Task Name"
+                            onChange={(e) =>
+                                setState({
+                                    ...state,
+                                    name: e.target.value,
+                                })
+                            }
+                        />
+                        <FormErrorMessage>Task name cannot be zero characters long.</FormErrorMessage>
```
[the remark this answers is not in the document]
oskarikerppo: changed in 66133e8

components/tasks/TaskTable.tsx
LeadingMoominExpert: I think so. It's much easier to read. 👍🏻 
Writing short helper functions like that is a good habit IMO. On top of that now there's a nice low hanging fruit to improve test coverage, since it would be possible to write a unit test for your new helper function. Though I don't think there's a module for running tests here yet. @robertsiipola should we look into adding `jest` or similar to actually test something in the code?

robertsiipola: > I think so. It's much easier to read. 👍🏻 Writing short helper functions like that is a good habit IMO. On top of that now there's a nice low hanging fruit to improve test coverage, since it would be possible to write a unit test for your new helper function. Though I don't think there's a module for running tests here yet. @robertsiipola should we look into adding `jest` or similar to actually test something in the code?

Yes `jest` for unit testing helper functions is fine. Apart from that I'd prefer something like Cypress/Selenium/Puppeteer to test the actual frontend. For example Chrome now integrates with Puppeteer, so that you can "record" a user flow from which the browser generates a javascript file for Puppeteer that keeps track of all the user/browser events. https://developer.chrome.com/docs/devtools/recorder/. 

Puppeteer also has a pretty decent integration with jest: https://jestjs.io/docs/puppeteer

oskarikerppo: Date checking moved to utils in fcdfd31

LeadingMoominExpert (approved):

PR #69, opened 2022-01-14
feat: added jest to epoc and wrote a test

oskarikerppo:


robertsiipola (approved):
Let's add these to Github Actions in a separate PR.

__tests__/checkDateOrder.test.ts
```diff
@@ -0,0 +1,11 @@
+import { expect, test } from '@jest/globals';
+import checkDateOrder from '../lib/utils/checkDateOrder';
+
+test('checkDateOrder returns true if second date precedes first date', () => {
+    const date1 = '2022-01-14';
+    const date2 = '2022-01-13';
+    expect(checkDateOrder(date1, date2)).toBe(true);
```
LeadingMoominExpert, line 7: Jest offers some better assertion checks, here you should use `.toBeTruthy()`

__tests__/checkDateOrder.test.ts
```diff
@@ -0,0 +1,11 @@
...
+    const date3 = '2022-01-14';
+    const date4 = '2022-01-15';
+    expect(checkDateOrder(date3, date4)).toBe(false);
```
LeadingMoominExpert, line 10: `.toBeFalsy()`

LeadingMoominExpert (commented):
The tests you run in the `checkDateOrder.test.ts` are two separate tests. You can run multiple tests in a single module by doing something like 

```typescript
describe('checkDateOrder returns true if second date precedes first date', () => {
    test('first date preceding second date should return true', () => {
        const date1 = '2022-01-14';
        const date2 = '2022-01-13';
        expect(checkDateOrder(date1, date2)).toBeTruthy();
    });

    test('second date preceding first date should return false', () => {
        const date3 = '2022-01-14';
        const date4 = '2022-01-15';
        expect(checkDateOrder(date3, date4)).toBeFalsy();
    });
});
```

LeadingMoominExpert:
> Let's add these to Github Actions in a separate PR.

Could be added alreaady? It's just one more line `yarn test` in the gha manifests? I did notice the file name is `.github/workflows/main.yaml` but the workflow name is `housekeeping`. filename should be changed.

oskarikerppo:
The tests were improved in 665e25d. However, there apparently was a conflict with babel and SWC. The project wouldn't compile correctly. This was fixed in e73ed5f by removing babel depencies and replacing jest with SWC version.

LeadingMoominExpert (approved):

PR #71, opened 2022-01-20
fix: fixed fetch bug on first render of id page

oskarikerppo:


oskarikerppo:
Network traffic upon refreshing http://localhost:3000/projects/1 before this fix
Screenshot from 2022-01-20 11-21-04
:

oskarikerppo:
After fix: 
Screenshot from 2022-01-20 11-21-41

oskarikerppo:
To clarify, upon refreshing page the `router.query` was undefined. This caused problems in the `useData` hook.

lib/utils/fetch.ts
```diff
@@ -1,10 +1,11 @@
 import { Auth } from 'aws-amplify';
 
 // https://eckertalex.dev/blog/typescript-fetch-wrapper
-async function http<T>(path: string, config?: RequestInit): Promise<T> {
+async function http<T>(path?: string, config?: RequestInit): Promise<T> {
```
LeadingMoominExpert, line 4: I'm not a huge fan of the path being nullable. Having a nullable function parameter here just adds complexity to the function which doesn't need it.

What's the case that requires converting path from undefined to empty string?

LeadingMoominExpert: Based on you screenshots there's no api queries ran to empty path?

oskarikerppo: > Based on you screenshots there's no api queries ran to empty path?

The first screenshot is from the main branch. Second screenshot there's not api queries with empty path as supposedly fetch catches the error with empty path

PR #72, opened 2022-01-21
fix: fixed first render bug with useData

oskarikerppo:


oskarikerppo:
This is another fix to the bug where first render has router.query return undefined.

PR #76, opened 2022-01-21
feat: billable status can be set from tasktable modal

oskarikerppo:

PR #74, opened 2022-01-21
feat: tasks can be archived by button press

oskarikerppo:

PR #78, opened 2022-01-21
refactor: major refactor

mapehe:
## Major refactor suggestion

**Objective:** Refactor the existing codebase to a significantly higher level of quality without altering the functionality.

At the moment, there are several issues with the state of the project as per the `main` branch. They vary from small issues with less than idiomatic code to somewhat critical design flaws. This is a rather large pull request, where I attempt to bring the codebase to a much more maintainable level without altering in any way how the app works. The files have been rearranged a little, but they should quite nicely correspond to those of the `main` branch. Since this is quite large PR, I'll try to explain my changes in plain text below.

### Major issue: Nullable component props

At the moment there are several components `<Component someProp={someProp} />` which make no sense with `someProp = undefined`. At the moment we set `someProp` nullable in `Component` props throughout. This is just asking for all sorts of nasty trouble: at some point someone will eventually mess up and break things. It also makes the render functions really hard to read since we have to do all sorts of iffing inside the render.

The correct way to design this is to set  `someProp` non-nullable and conditionally render in the parent component i.e.
```
someProp && <Component someProp={someProp}/> 
```
The same holds for [parameters that hooks depend on](https://github.com/three-consulting/epoc-frontend/blob/refactor/major-refactor/pages/projects/%5Bid%5D/index.tsx#L103-L107).

### Major issue: No proper form validation, incorrect DTO types

Form fields are not validated correctly against non-nullability. While some fields are marked [required](https://github.com/three-consulting/epoc-frontend/blob/eb996de54ffce86d275f8759aa5521ec06d1bfcd/components/projects/NewProject/ProjectForm.tsx#L99), this should be regarded as a UX feature that prevents the user from running into weird UI errors when some required fields are left empty. An additional layer of validation on the typescript level must be added so that [there is a type that represents a form with possibly missing fields, that is later validated against non-nullability rules](https://github.com/three-consulting/epoc-frontend/blob/refactor/major-refactor/components/form/ProjectForm.tsx#L10-L26).

As per `main` the api types are incorrect as some required fields are marked nullable. I've updated [the types](https://github.com/three-consulting/epoc-frontend/blob/refactor/major-refactor/lib/types/api.ts#L102-L112) so that the required fields are reflected in the generated types. I've also added a corresponding [backend PR](https://github.com/three-consulting/epoc-backend/pull/78).

### Major issue: useData

The current `useData` is too inflexible to be used properly. I ​think `useData` should be [rewritten](https://github.com/three-consulting/epoc-frontend/blob/refactor/major-refactor/lib/hooks/useData.tsx) to allow this pattern
```
    const projectRequest = useCallback(() => listProjects(), []);
    const [projectResponse] = useData(projectRequest);
```
i.e. a general API callback that may or may not have additional parameters such as `id` set (see [example](https://github.com/three-consulting/epoc-frontend/pull/78/files#diff-3214e52630eb13ec941ccf4b21156845c4ecc83e1381a6fd4acf67cb6c4a1775R23) of setting a paramter).

Also the return value should be properly typed: `data` is defined if and only if `isSuccess` is `true`. This makes for more idiomatic code since the correct way to conditionally render becomes
```
someResponse.isSuccess && <Component someProp={someResponse.data}/> 
```
This solves many issues in relation to nullable fields. The newly suggested `useData` returns a pending request and a refresh function [example](https://github.com/three-consulting/epoc-frontend/blob/refactor/major-refactor/pages/projects/new.tsx#L14)).
```
    const customerRequest = useCallback(() => listCustomers(), []);
    const [customerResponse, refreshCustomers] = useData(customerRequest);
```
`refreshCustomers` can then be used in some component below to fetch customer data again

I suggest we try to [work with promises directly](https://github.com/three-consulting/epoc-frontend/blob/refactor/major-refactor/lib/utils/fetch.ts#L5-L20) and enjoy the fine grained control we get over the way things are refreshed by `useMemo` dep changes. If this turns out to be too difficult, we can always switch back to some library.
 
### Major issue: Unsafe API calls from weird states

At the moment each component depends on a single state of `StateType` with entries that form an object type that doesn't really mean anything in terms of DTOs ([example](https://github.com/three-consulting/epoc-frontend/blob/eb996de54ffce86d275f8759aa5521ec06d1bfcd/components/tasks/TaskTable.tsx#L31-L38)). These states are then used to make [API calls](https://github.com/three-consulting/epoc-frontend/blob/eb996de54ffce86d275f8759aa5521ec06d1bfcd/components/tasks/TaskTable.tsx#L74) from within the components to manually manipulated URLs. IMO this is just asking for a lot of trouble in the future: API calls should only be made from within typed functions that are [separated](https://github.com/three-consulting/epoc-frontend/blob/refactor/major-refactor/lib/utils/apiRequests.ts) from components and the URLs should never be manipulated manually in components.

Component states should be idiomatic: We should use several `useState` hooks where each state means something important such as `ProjectDTO`, which can then naturally be passed around. The current design is just a lot of extra work.

### Major issue: Non idiomatic code

The `main` branch is really hard to read and it took me quite a while to figure out what the components are supposed to do. I've spent quite a bit of time attempting to redesign the components so that their intent is as clear as possible. API fetch calls, for example, should not be made in some  component way below the main page.

As per `main` branch there is no clear separation of duties: `TaskTable` is simultaneously a Table and a Form that performs Ajax requests. It's much better to split components by functionality and simplify the individual component states to something more manageable.
​
### Minor issue: Style

There are various style things I've commented on below.

Closes #72 #74 #76


components/form/ProjectForm.tsx
```diff
@@ -0,0 +1,187 @@
...
+    const [formStatus, setFormStatus] = useState<FormStatus>(FormStatus.IDLE);
+    const router = useRouter();
+
+    const handleCustomerChange = (e: React.FormEvent<HTMLSelectElement>) => {
+        e.preventDefault();
+        const id = e.currentTarget.value;
+        if (id) {
+            const customer = customers.find((el) => el.id === Number(id));
+            setProject({ ...project, customer: customer });
+        }
```
mapehe, line 46: This doesn't work, but it produces the same request as `main` branch. Need to fix.

components/form/CustomerForm.tsx
```diff
@@ -14,54 +14,36 @@ import {
...
-import { components } from '@/lib/types/api';
 import * as fetch from '@/lib/utils/fetch';
-import { FormStatus } from './reducer';
 import { useSWRConfig } from 'swr';
 import ErrorAlert from '@/components/common/ErrorAlert';
+import { FormStatus } from './ProjectForm';
+import { CustomerDTO } from '@/lib/types/dto';
+import { customerEndpointURL } from '@/lib/const';
 
-type StateType = {
```
mapehe, line 23: Usually there is no reason to prefer the use of single `useState` instead of multiple `useState` hooks where the values actually mean something. For example, in this case, it's much clearer to split this into a `CustomerDTO` state and separate states that are related to the form state.

components/table/TimesheetTable.tsx
```diff
@@ -133,51 +90,52 @@ function TimesheetTable({ timesheets, project }: TimesheetTableProps): JSX.Eleme
...
+                                            <Td>{timesheet.allocation} %</Td>
                                             <Td>
-                                                <Button onClick={(e) => archiveTimesheet(el, e)}>x</Button>
+                                                <Button onClick={(e) => archiveTimesheet(timesheet, e)}>x</Button>
                                             </Td>
                                         </Tr>
                                     );
                                 }
                             })}
-                            {state.formStatus == 'ERROR' ? <ErrorAlert></ErrorAlert> : <Box></Box>}
```
mapehe, line 179: A standard way to return
```
<ErrorAlert/>
<Box>{state.errorMessage}</Box>
```
is to wrap it into a fragment:
```
<>
    <ErrorAlert/>
   <Box>{state.errorMessage}</Box>
</>
```

mapehe: The standard way to conditionally return nothing is 
```
condition ? <Component /> : null
```

mapehe:
```
<Component></Component>
```
equals
```
<Component/>
```
the latter should be preferred

components/projects/NewProject/reducer.ts
```diff
@@ -1,80 +0,0 @@
-import { components } from '@/lib/types/api';
```
mapehe, line 1: I don't see the point in any of this so I'm just gonna suggest we remove it

components/table/ProjectTable.tsx
```diff
@@ -1,11 +1,11 @@
 import React from 'react';
 import { Box, Heading } from '@chakra-ui/layout';
 import { Table, TableCaption, Thead, Tr, Td, Th, Tbody } from '@chakra-ui/react';
-import { components } from '@/lib/types/api';
 import Link from 'next/link';
+import { ProjectDTO } from '@/lib/types/dto';
 
 type ProjectListProps = {
-    projects: components['schemas']['ProjectDTO'][];
```
mapehe, line 8: There does not seem to be a good reason to prefer `components['schemas']['ProjectDTO'][]` over `ProjectDTO[]` with an exported conversion somewhere

components/table/TimesheetTable.tsx
```diff
@@ -18,108 +17,66 @@ import {
...
 import * as fetch from '@/lib/utils/fetch';
 import ErrorAlert from '../common/ErrorAlert';
 import useData from '@/lib/hooks/useData';
 import Loading from '../common/Loading';
+import { FormStatus } from '../form/ProjectForm';
+import { employeeEndpointURL, timesheetEndpointURL } from '@/lib/const';
+import { TimesheetDTO, ProjectDTO, EmployeeDTO } from '@/lib/types/dto';
 
 type TimesheetTableProps = {
-    timesheets?: components['schemas']['TimesheetDTO'][];
```
mapehe, line 28: There is probably not a good reason for this to be nullable. Preferring conditional render
```
p && <Component p={p}/>
```
over a nullable `p` will almost certainly prevent quite a few future bugs

components/form/ProjectForm.tsx
mapehe: As a bonus, we reduce [clutter](https://github.com/three-consulting/epoc-frontend/pull/78/files#diff-0371e3443d5f2360efcc6f352cb9af86177163d97e5dcaf6fa718addc441eb26L63-L69)

pages/projects/[id]/edit.tsx
```diff
@@ -2,44 +2,52 @@ import React from 'react';
...
-            {isError && <ErrorAlert title={isError.name} message={isError.name}></ErrorAlert>}
-            <ProjectForm customers={customers} employees={employees} method="PUT" project={projects}></ProjectForm>
         </Layout>
     );
+}
+
+const Edit: NextPage = () => {
+    const router = useRouter();
+    const { id } = router.query;
+    return id ? <EditProjectForm id={id[0]} /> : null;
```
Comment (mapehe) [inline pages/projects/[id]/edit.tsx:50]: The correct way to handle the possibly missing `id` that's used in hooks is to conditionally render iff `id` is defined.

Comment (mapehe) [reply pages/projects/[id]/edit.tsx:50]: Notice that `EditProjectForm` requires `id` to be defined. Thus the hooks work nicely.

components/form/CustomerForm.tsx
mapehe: As a bonus, we reduce [clutter](https://github.com/three-consulting/epoc-frontend/pull/78/files#diff-0371e3443d5f2360efcc6f352cb9af86177163d97e5dcaf6fa718addc441eb26L63-L69)

components/table/ProjectTable.tsx
```diff
@@ -21,12 +21,12 @@ function ProjectTable({ projects }: ProjectListProps): JSX.Element {
                         </Tr>
                     </Thead>
                     <Tbody>
-                        {projects.map((el, idx) => {
```
mapehe, line 24: `el` is a non-idiomatic varibale name, something like `project` is much easier to read without context and should be preferred

lib/utils/fetch.ts
```diff
@@ -1,38 +1,52 @@
...
+async function http<T>(path: string, config?: RequestInit): Promise<ResponseWithStatus<T>> {
     const authSession = await Auth.currentSession();
     const jwt = authSession.getIdToken()?.getJwtToken();
     const request = new Request(path, {
         headers: { 'Content-Type': 'application/json', Authorization: `Bearer ${jwt}` },
         ...config,
     });
-    const response = await fetch(request);
-
-    if (!response.ok) {
```
mapehe, line 13: It's probably not a good idea _not_ to clearly communicate this so that it's accessible from the components. The `data` field is unreliable since .e.g `!""==true`

__tests__/checkDateOrder.test.ts
```diff
@@ -1,20 +0,0 @@
-import { expect, test } from '@jest/globals';
```
mapehe, line 1: If we think, we need a separate util function for something like this, we should definitely not write it ourselves. Something like lodash should be preferred.

components/form/ProjectForm.tsx
mapehe: Fixed: This was a bug due to incorrect API types

yarn.lock
```diff
@@ -8998,13 +8993,6 @@ supports-hyperlinks@^2.0.0:
     has-flag "^4.0.0"
     supports-color "^7.0.0"
 
-swr@^1.0.1:
```
mapehe, line 9001: I don't understand why we need swr. We can reverse this removal if there is an actual reason for us to use it.

components/table/TaskTable.tsx
```diff
@@ -0,0 +1,157 @@
...
+    }
+};
+
+interface TaskTableProps {
+    project: ProjectDTO;
+    tasks: TaskDTO[];
+}
+
+function TaskTable({ project, tasks }: TaskTableProps): JSX.Element {
+    const [taskFields, setTaskFields] = useState<TaskFields>({
```
mapehe, line 44: The idiomatic way to handle forms like these is to use a state with nullable fields representing a form that is still being filled and validate upon submit

components/form/ProjectForm.tsx
```diff
@@ -0,0 +1,195 @@
...
+    }
+};
+
+type ProjectFormPropsBase = {
+    employees: EmployeeDTO[];
+    customers: CustomerDTO[];
+    refreshCustomers: () => void;
+};
+
+type ProjectFormProps =
```
mapehe, line 34: The type `ProjectFormProps` should reflect the fact that `project` is defined iff `method=put` and is `undefined` otherwise.

components/tasks/TaskTable.tsx
```diff
@@ -1,222 +0,0 @@
...
-        errorMessage: '',
-    });
-
-    const {
-        data: tasks,
-        isError: taskError,
-        isLoading: tasksLoading,
-    } = useData<components['schemas']['TaskDTO'][]>('task', { projectId: `${project?.id}` });
-
-    const url = `${process.env.NEXT_PUBLIC_API_URL}/task`;
```
mapehe, line 56: Definition of config parameters in components is a red flag.

components/tasks/TaskTable.tsx
```diff
@@ -1,222 +0,0 @@
...
-        startDate: new Date().toISOString().split('T')[0],
-        formStatus: FormStatus.IDLE,
-        errorMessage: '',
-    });
-
-    const {
-        data: tasks,
-        isError: taskError,
-        isLoading: tasksLoading,
-    } = useData<components['schemas']['TaskDTO'][]>('task', { projectId: `${project?.id}` });
```
mapehe, line 54: I don't think it's a good idea to make fetch requests in these kinds of components. I suggest that, as a rule of a thumb, anything in the `components` directory should not make fetch calls. Fetch calls should be made by components in `pages` directory and passed downwards as arguments in conditional render statements.

components/tasks/TaskTable.tsx
```diff
@@ -1,222 +0,0 @@
...
-        <Flex
-            flexDirection="column"
-            backgroundColor="white"
-            border="1px solid"
-            borderColor="gray.400"
-            borderRadius="0.2rem"
-            padding="1rem 1rem"
-            marginTop="1.5rem"
-        >
-            {tasksLoading && <Loading></Loading>}
```
mapehe, line 100: These kinds of conditional renders add unnecessary clutter into these components. Should fetch data higher up and pass it downwards as props.

components/tasks/TaskTable.tsx
```diff
@@ -1,222 +0,0 @@
...
-        const createTaskRequest: components['schemas']['TaskDTO'] = {
-            name: state.name,
-            description: state.description,
-            startDate: state.startDate,
-            endDate: state.endDate,
-            project: project,
-        };
-        setState({
-            ...state,
-            formStatus: FormStatus.LOADING,
```
mapehe, line 71: I don't understand why we need this `formStatus` thing. I'm just going to suggest we remove it throughout and render ´<Error>` components iff `errorMessage` is defined.

lib/utils/apiRequests.ts
```diff
@@ -0,0 +1,34 @@
...
+const taskEndpointURL = `${process.env.NEXT_PUBLIC_API_URL}/task`;
+const employeeEndpointURL = `${process.env.NEXT_PUBLIC_API_URL}/employee`;
+const timesheetEndpointURL = `${process.env.NEXT_PUBLIC_API_URL}/timesheet`;
+const projectIdEndpointURL = (id: number): string => `${process.env.NEXT_PUBLIC_API_URL}/project/${id}`;
+
+export const getProject = (projectId: number): Promise<ResponseWithStatus<ProjectDTO>> =>
+    get<ProjectDTO>(projectIdEndpointURL(projectId));
+export const listProjects = (): Promise<ResponseWithStatus<ProjectDTO[]>> => get<ProjectDTO[]>(projectEndpointURL);
+
+export const listCustomers = (): Promise<ResponseWithStatus<CustomerDTO[]>> => get<CustomerDTO[]>(customerEndpointURL);
```
mapehe, line 15: These type signatures suck: I'd like to define
```
export type ApiRequest<T> = Promise<ResponseWithStatus<T>>
```
but this doesn't work due to [a bug](https://github.com/microsoft/TypeScript/issues/27987) in typescript itself.

lib/utils/fetch.ts
```diff
@@ -1,38 +1,59 @@
 import { Auth } from 'aws-amplify';
 
-// https://eckertalex.dev/blog/typescript-fetch-wrapper
-async function http<T>(path: string, config?: RequestInit): Promise<T> {
+export type ResponseWithStatus<T> = { data: T; isSuccess: true } | { isSuccess: false; errorMessage: string };
```
mapehe, line 3: I've redesigned this. I don't think it's a good idea to hide the asynchronous nature of requests into SWR since we lose control over the way promises are handled.

pages/projects/[id]/index.tsx
```diff
@@ -1,115 +1,114 @@
...
-    const router = useRouter();
-    const { id } = router.query;
-    const { data: project, isError, isLoading } = useData<components['schemas']['ProjectDTO']>(`project/${id}`);
-    const { isOpen, onOpen, onClose } = useDisclosure();
-    const [state, setState] = useState<StateType>({
-        formStatus: FormStatus.IDLE,
-        errorMessage: '',
-    });
-    const url = `${process.env.NEXT_PUBLIC_API_URL}/project`;
-    const { mutate } = useSWRConfig();
```
Comment (mapehe) [inline pages/projects/[id]/index.tsx:34]: I don't understand why we need this

lib/hooks/useData.tsx
```diff
@@ -1,18 +1,46 @@
...
-import * as fetch from '../utils/fetch';
+import { useMemo, useState } from 'react';
+import { ResponseWithStatus } from '../utils/fetch';
 
-type DataResponse<T> = {
-    data?: T;
-    isLoading: boolean;
-    isError?: Error;
-};
+type UseDataType<T> =
```
mapehe, line 4: The correct type for possibly failing requests is something that has separate `isSuccess` and `data` fields.

lib/utils/apiRequests.ts
```diff
@@ -0,0 +1,34 @@
+import { CustomerDTO, EmployeeDTO, ProjectDTO, TaskDTO, TimesheetDTO } from '../types/dto';
+import { get, post, put, ResponseWithStatus } from './fetch';
+
+const projectEndpointURL = `${process.env.NEXT_PUBLIC_API_URL}/project`;
```
mapehe, line 4: URLS must be set in a centralized way and never manipulated manually afterwards. Would it be a good idea if these came from the openapi docs directly?

components/table/TaskTable.tsx
```diff
@@ -0,0 +1,75 @@
...
+                <Button colorScheme="blue" onClick={() => setDisplayNewTaskForm(true)}>
+                    Add Task
+                </Button>
+            </Flex>
+            <Modal isOpen={displayNewTaskForm} onClose={() => setDisplayNewTaskForm(false)}>
+                <ModalOverlay />
+                <ModalContent px="0.5rem">
+                    <ModalHeader>Add task to project</ModalHeader>
+                    <ModalCloseButton />
+                    <TaskForm project={project} onClose={() => setDisplayNewTaskForm(false)} />
```
mapehe, line 68: I think it's not a great idea to not define this form separately, so I'm gonna suggest that

components/table/TaskTable.tsx
```diff
@@ -0,0 +1,75 @@
...
+import React, { useState } from 'react';
+import { TaskForm } from '../form/TaskForm';
+
+interface TaskTableProps {
+    project: ProjectDTO;
+    tasks: TaskDTO[];
+}
+
+function TaskTable({ project, tasks }: TaskTableProps): JSX.Element {
+    const [displayNewTaskForm, setDisplayNewTaskForm] = useState(false);
```
mapehe, line 15: I think we're better off with just a simple boolean state instead of some fancy chakra stuff.

lib/hooks/useData.tsx
```diff
@@ -1,18 +1,46 @@
...
+          isLoading: true;
+          isError: false;
+      };
 
-function useData<T>(endpoint: string, queryParams?: Record<string, string>): DataResponse<T> {
-    const url = new URL(endpoint, process.env.NEXT_PUBLIC_API_URL);
-    url.search = new URLSearchParams(queryParams).toString();
+function useDataInternalState<T>(request: Promise<ResponseWithStatus<T>>): UseDataType<T> {
+    const [response, setResponse] = useState<ResponseWithStatus<T>>();
+    request.then(setResponse);
```
mapehe, line 25: I don't think we need SWR mutate anymore because this does the same thing

robertsiipola (changes requested):
# Why use a separate library to handle server state

The reason to use an async state manager for server data (or caching data) is because server state is inherently different from UI state and having dedicated tools for it lessens the burden on the developer. 

While we could write our own logic with useEffect and useState to handle server state and functionality like cache revalidation, its a lot of work to get it right (plus maintenance) and the opportunity cost seems high when you have libraries like SWR and [React Query](https://react-query.tanstack.com/) (RQ) which were meant to solve these problems, are reaching maturity and are tested in production. They both use a caching control strategy called [stale-while-revalidate](https://web.dev/stale-while-revalidate/), which is now an extension for the HTTP Cache-Control header.

Using said libraries also reduces the need for a global state solution, and we can get by with useState (or useReducer) for all of our UI state needs. To quote [Kent C. Dodds](https://kentcdodds.com/blog/application-state-management-with-react):

> One last thing I want to add. There are various categories of state, but every type of state can fall into one of two buckets:
>
>1. Server Cache - State that's actually stored on the server and we store in the client for quick-access (like user data).
> 2. UI State - State that's only useful in the UI for controlling the interactive parts of our app (like modal isOpen state).
>
> We make a mistake when we combine the two. Server cache has inherently different problems from UI state and therefore needs to be managed differently. If you embrace the fact that what you have is not actually state at all but is instead a cache of state, then you can start thinking about it correctly and therefore managing it correctly.
>
>You can definitely manage this yourself with your own useState or useReducer with the right useContext here and there. But allow me to help you cut to the chase and say that caching is a really hard problem (some say that it's one of the hardest in computer science) and it would be wise to stand on the shoulders of giants on this one.
>
>This is why I use and recommend react-query for this kind of state. I know I know, I told you that you don't need a state management library, but I don't really consider react-query to be a state management library. I consider it to be a cache. And it's a darn good one. Give it a look! That Tanner Linsley is one smart cookie.

That said, there is probably a case to be made for using RQ over SWR, because RQ is more popular, has better documentation and has its own dev tools.

So either revert the SWR changes or swap SWR for React Query.

Other than that these changes look good to me.

__tests__/test.test.ts
```diff
@@ -0,0 +1,5 @@
+import { test } from '@jest/globals';
+
+describe('trivial test', () => {
+    test('trivial test', async () => true);
+});
```
LeadingMoominExpert, line 5: You can just remove `__tests__` if there's no actual tests

components/form/ProjectForm.tsx
```diff
@@ -0,0 +1,184 @@
...
+        e.preventDefault();
+        const id = e.currentTarget.value;
+        if (id) {
+            const employee = employees.find((el) => el.id === Number(id));
+            setProjectFields({ ...projectFields, managingEmployee: employee });
+        }
+    };
+
+    const invalidEndDate =
+        (projectFields.startDate && projectFields.endDate && projectFields.startDate > projectFields.endDate) || false;
```
LeadingMoominExpert, line 49: This is altered behavior because `endDate` being `undefined` is a valid case.

LeadingMoominExpert: Hence why it was written as a separate function so there's no need to use nested ternary expressions

aleksimyllyoja (approved):
👍

lib/types/dto.ts
```diff
@@ -0,0 +1,7 @@
+import { components } from './api';
+
+export type ProjectDTO = components['schemas']['ProjectDTO'];
+export type EmployeeDTO = components['schemas']['EmployeeDTO'];
+export type CustomerDTO = components['schemas']['CustomerDTO'];
+export type TaskDTO = components['schemas']['TaskDTO'];
+export type TimesheetDTO = components['schemas']['TimesheetDTO'];
```
LeadingMoominExpert, line 7: nice

lib/utils/fetch.ts
```diff
@@ -18,17 +17,26 @@ async function http<T>(path: string, config?: RequestInit): Promise<T> {
...
+    return await http<T>(url.href, init);
 }
 
-export async function post<T, U>(path: string, body: T, config?: RequestInit): Promise<U> {
+export async function post<T, U>(
+    path: string,
+    body: T,
+    params?: Record<string, string | number>,
+    config?: RequestInit,
+): Promise<U> {
```
LeadingMoominExpert, line 32: You could also spread the parameters on their respective rows like this in the other functions in the file too, to improve readability.

package.json
```diff
@@ -29,7 +29,7 @@
         "react-dom": "17.0.2",
         "react-icons": "^4.3.1",
         "react-responsive": "^9.0.0-beta.4",
-        "swr": "^1.0.1"
+        "swr": "^1.1.2"
```
LeadingMoominExpert, line 32: While doing a major refactoring already could you update other deps as well? Dependabot has been crying about the used `next` for a while now

LeadingMoominExpert (changes requested):

components/form/ProjectForm.tsx
mapehe: Can an undefined `endDate` be invalid? If `endDate` is undefined this expression results in false.

lib/utils/fetch.ts
mapehe: The linter won't let me 😢

__tests__/test.test.ts
mapehe: In that case `yarn test` will return 1. That probably requires some changes to the GH tasks?

LeadingMoominExpert: Remove this line: https://github.com/three-consulting/epoc-frontend/blob/refactor/major-refactor/.github/workflows/housekeeping.yml#L29

robertsiipola (approved):

components/form/ProjectForm.tsx
LeadingMoominExpert: The logic is that endDate doesn't need to be defined, but we should always know at least when project or anything else has started
https://github.com/three-consulting/epoc-backend/blob/main/src/main/kotlin/three/consulting/epoc/dto/ProjectDTO.kt#L13-L14
The service layer checks if `endDate` is not null, that it's not before `startDate`. So right now null `endDate` is perfectly fine.

mapehe: So `endDate` is `null`, here we get `invalidEndDate === false` as intended?

lib/types/dto.ts
LeadingMoominExpert: I'm also thinking that you could probably drop the `DTO` from the exported type names? They just come from generated types from the backend because there's a need to differentiate between dao's and dto's

components/form/ProjectForm.tsx
LeadingMoominExpert: AAAAAAAA it's INvalid. I don't know how to read. I read `isValidEndDate`

LeadingMoominExpert:
Proper upgrades weren't proper and the updated tsconfigs are not passing through checks

LeadingMoominExpert (approved):
jee

PR #81, opened 2022-01-25
Refactor/idiomatic forms

mapehe:
The current `ProjectForm` actually wants to be two separate forms: The expression
```
<CreateProjectForm customers={customers} employees=(employees} />
```
makes much more sense without context than
```
<ProjectForm customers={customers} employees=(employees} project={undefined} projectId={undefined} />
```
I suggest all forms doing the basic CRUD should be `Create[X]Form` or `Edit[X]Form` accordingly. Refactored the project form and renamed the others.

### DIFF
--- a/components/form/CustomerForm.tsx
+++ b/components/form/CustomerForm.tsx
@@ -29,7 +29,7 @@ const validateCustomerFields = (fields: CustomerFields): Customer => {
     }
 };
 
-function CustomerForm(): JSX.Element {
+function CreateCustomerForm(): JSX.Element {
     const [customerFields, setCustomerFields] = useState<CustomerFields>({});
     const { isOpen, onOpen, onClose } = useDisclosure();
     const [errorMessage, setErrorMessage] = useState<string>('');
@@ -101,4 +101,4 @@ function CustomerForm(): JSX.Element {
     );
 }
 
-export default CustomerForm;
+export default CreateCustomerForm;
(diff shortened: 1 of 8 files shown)


components/form/ProjectForm.tsx
```diff
@@ -147,37 +149,32 @@ function ProjectFormBody({ project: projectOrNull, customers, employees, onSubmi
     );
 }
 
-type ProjectFormPropsBase = {
+type CreateProjectFormProps = {
     employees: Employee[];
     customers: Customer[];
 };
 
-type CreateProjectFormProps = ProjectFormPropsBase & { project: undefined; projectId: undefined };
```
mapehe, line 155: As a bonus, we get rid of this complex type logic.

oskarikerppo (approved):

components/form/CustomerForm.tsx
```diff
@@ -29,7 +29,7 @@ const validateCustomerFields = (fields: CustomerFields): Customer => {
     }
 };
 
-function CustomerForm(): JSX.Element {
+function CreateCustomerForm(): JSX.Element {
```
LeadingMoominExpert, line 32: Change the filename too?

But this naming convention is better. `CustomerForm` doesn't really tell what the form is for.

LeadingMoominExpert (approved):

PR #82, opened 2022-01-25
refactor: parent agnostic forms

mapehe:
Ideally it should be possible to render the forms `CreateClientForm`, `CreateTaskForm` and `CreateTimesheetForm` anywhere: They should have no idea whether the parent is a modal or something else. This promotes component reuse and makes these forms much more useful in the long term.

I suggest we introduce optional `afterSubmit` and `onCancel` functions to forms. They allow patterns like
```
<CreateTaskForm afterSubmit={ closeModal } onCancel={ closeModal } />
```
in the parent.

A future improvement is that `afterSubmit` returns the created item to allow `afterSubmit( ( createdItem ) => doSomething(createdItem))`. You could, for example, redirect to the URL of a newly created project with
```
afterSubmit((newlyCreatedProject) => router.push(`/projects/${newlyCreatedProject.id}`))
```
this will not be done here.

### DIFF
--- a/components/form/TaskForm.tsx
+++ b/components/form/TaskForm.tsx
@@ -1,16 +1,18 @@
 import { useUpdateTasks } from '@/lib/hooks/useTasks';
 import { Project, Task } from '@/lib/types/apiTypes';
-import { FormControl, FormLabel, Input, FormErrorMessage, ModalFooter, Button, Box } from '@chakra-ui/react';
+import { FormBase } from '@/lib/types/forms';
+import { FormControl, FormLabel, Input, FormErrorMessage, Button, Box } from '@chakra-ui/react';
 import React, { useState } from 'react';
 import ErrorAlert from '../common/ErrorAlert';
 
 type TaskFields = Partial<Task> & { project: Project };
 
-const validateTaskFields = (fields: TaskFields): Task => {
-    const { name } = fields;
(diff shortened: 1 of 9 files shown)


components/form/TaskForm.tsx
```diff
@@ -1,30 +1,31 @@
...
 type TaskFields = Partial<Task> & { project: Project };
 
-const validateTaskFields = (fields: TaskFields): Task => {
-    const { name } = fields;
+const validateTaskFields = (fields: TaskFields, projectId: number): Task => {
+    const { name, project } = fields;
     if (name) {
         return {
             ...fields,
+            project: { ...project, id: projectId },
```
mapehe, line 15: This is important BTW, a missing `projectId` could result in some really nasty hard-to-find bugs

components/table/TaskTable.tsx
```diff
@@ -80,7 +80,7 @@ function TaskTable({ project, tasks }: TaskTableProps): JSX.Element {
                             <CreateTaskForm
                                 project={project}
                                 projectId={project.id}
-                                onClose={() => setDisplayNewTaskForm(false)}
```
oskarikerppo, line 83: Missing onCancel

mapehe:
Closing this in favour of #86

PR #85, opened 2022-01-25
feat: mutate detail cache keys on update

mapehe:
## Implement match mutate

@oskarikerppo noticed a missing feature in the current hooks: If we perform `putProject`, any components that use data from `useProjectDetail` are not updated. This is because we don't mutate the correct cache keys in `useUpdateProject`-hook.

What seems to be the correct way to achieve this is to use [match mutate](https://github.com/three-consulting/epoc-frontend/security/dependabot/yarn.lock/node-fetch/open), to mutate all API keys that match a RegEx.

### DIFF
--- a/lib/utils/matchMutate.js
+++ b/lib/utils/matchMutate.js
@@ -1,0 +1,21 @@
+import { useSWRConfig } from 'swr';
+
+export const useMatchMutate = () => {
+    const { cache, mutate } = useSWRConfig();
+    return (matcher, ...args) => {
+        if (!(cache instanceof Map)) {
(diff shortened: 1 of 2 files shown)


lib/utils/matchMutate.js
```diff
@@ -0,0 +1,21 @@
+import { useSWRConfig } from 'swr';
```
mapehe, line 1: This is just copypasta from https://swr.vercel.app/docs/advanced/cache#mutate-multiple-keys-from-regex

oskarikerppo (approved):

PR #88, opened 2022-01-27
feat: router pushes to /id upon creating a project

oskarikerppo:


pages/projects/new.tsx
```diff
@@ -34,7 +38,7 @@ const New: NextPage = () => {
                 <CreateProjectForm
                     customers={customersResponse.data}
                     employees={employeesResponse.data}
-                    afterSubmit={redirectToProjectList}
+                    afterSubmit={(newProject: ApiUpdateResponse<Project>) => redirectToProjectDetails(newProject)}
```
mapehe, line 41:
```
afterSubmit={(newProject: ApiUpdateResponse<Project>) => redirectToProjectDetails(newProject)}
```
is equal to
```
afterSubmit={redirectToProjectDetails}
```

pages/projects/new.tsx
```diff
@@ -20,6 +22,8 @@ const New: NextPage = () => {
         '';
 
     const redirectToProjectList = () => router.push('/projects');
+    const redirectToProjectDetails = (newProject: ApiUpdateResponse<Project>) =>
+        newProject.isSuccess && router.push(`/projects/${newProject.data.id}`);
```
mapehe, line 26: I'd prefer calling this `newProject` something like `projectCreate` instead, because the actual project is contained in the `data` field upon success.

components/form/TaskForm.tsx
```diff
@@ -8,12 +8,13 @@ import ErrorAlert from '../common/ErrorAlert';
 type TaskFields = Partial<Task> & { project: Project };
 
 const validateTaskFields = (fields: TaskFields, projectId: number): Task => {
-    const { name, project } = fields;
-    if (name) {
+    const { name, project, billable } = fields;
```
mapehe, line 11: I think this does not belong in this PR.

lib/types/api.ts
```diff
@@ -71,86 +71,120 @@ export interface paths {
 export interface components {
   schemas: {
     CustomerDTO: {
+      /** Format: int64 */
```
mapehe, line 74: I think these should be a separate PR.

mapehe (changes requested):

mapehe (approved):

PR #89, opened 2022-01-27
Feat/billable field to createtaskform

oskarikerppo:
Depends on  #88

### DIFF
--- a/components/form/TaskForm.tsx
+++ b/components/form/TaskForm.tsx
@@ -1,19 +1,20 @@
 import { useUpdateTasks } from '@/lib/hooks/useTasks';
 import { Project, Task } from '@/lib/types/apiTypes';
 import { FormBase } from '@/lib/types/forms';
-import { FormControl, FormLabel, Input, FormErrorMessage, Button, Box } from '@chakra-ui/react';
+import { FormControl, FormLabel, Input, FormErrorMessage, Button, Box, Flex, Checkbox } from '@chakra-ui/react';
 import React, { useState } from 'react';
 import ErrorAlert from '../common/ErrorAlert';
 
-type TaskFields = Partial<Task> & { project: Project };
+type TaskFields = Partial<Task> & { project: Project; billable: boolean };
 
 const validateTaskFields = (fields: TaskFields, projectId: number): Task => {
-    const { name, project } = fields;
-    if (name) {
+    const { name, project, billable } = fields;
+    if (name && billable !== undefined) {
         return {
             ...fields,
             project: { ...project, id: projectId },
             name,
+            billable,
         };
     } else {
         throw 'Invalid task form: missing required fields';
@@ -28,6 +29,7 @@ type TaskFormProps = FormBase<Task> & {
 export function CreateTaskForm({ project, projectId, afterSubmit, onCancel }: TaskFormProps): JSX.Element {
     const [taskFields, setTaskFields] = useState<TaskFields>({
         project: project,
+        billable: true,
     });
     const { postTask } = useUpdateTasks();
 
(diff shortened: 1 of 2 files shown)


mapehe (changes requested):
call

mapehe (approved):
LGTM

components/form/TaskForm.tsx
```diff
@@ -73,6 +74,20 @@ export function CreateTaskForm({ project, projectId, afterSubmit, onCancel }: Ta
                         }
                     />
                 </FormControl>
+                <FormControl py="0.5rem">
+                    <Flex flexDirection={'row'} alignItems={'baseline'}>
+                        <FormLabel>Billable</FormLabel>
+                        <Checkbox
+                            defaultChecked={true}
+                            onChange={(e) =>
```
mapehe, line 82: This doesn't work

mapehe (changes requested):

mapehe:
The form is [initialized](https://github.com/three-consulting/epoc-frontend/blob/1543acf10e3b6ab76b5444c4b6f52f467b67dd2b/components/form/TaskForm.tsx#L30-L32) incorrectly. In
```typescript
useState<TaskFields>({
        project: project,
    });
```
the object that is given as an argument to `useState<TaskFields>` represents the default state of the form. If you want the `billable` field to be checked by default, set
```
useState<TaskFields>({
        project: project,
        billable: true
    });
```

components/form/TaskForm.tsx
```diff
@@ -1,19 +1,20 @@
...
 import React, { useState } from 'react';
 import ErrorAlert from '../common/ErrorAlert';
 
 type TaskFields = Partial<Task> & { project: Project };
 
 const validateTaskFields = (fields: TaskFields, projectId: number): Task => {
-    const { name, project } = fields;
-    if (name) {
+    const { name, project, billable } = fields;
+    if (name && billable) {
```
mapehe, line 12: This always fails when `billable` is `false`. Should have `billable !== undefined`

components/form/TaskForm.tsx
```diff
@@ -1,19 +1,20 @@
 import { useUpdateTasks } from '@/lib/hooks/useTasks';
 import { Project, Task } from '@/lib/types/apiTypes';
 import { FormBase } from '@/lib/types/forms';
-import { FormControl, FormLabel, Input, FormErrorMessage, Button, Box } from '@chakra-ui/react';
+import { FormControl, FormLabel, Input, FormErrorMessage, Button, Box, Flex, Checkbox } from '@chakra-ui/react';
 import React, { useState } from 'react';
 import ErrorAlert from '../common/ErrorAlert';
 
 type TaskFields = Partial<Task> & { project: Project };
```
mapehe, line 8: It would probably be a good idea to edit this type to
```
type TaskFields = Partial<Task> & { project: Project, billable: boolean };
```

mapehe: This type represents the minimum information information that the initial state of a form must have. If we add `billable` field, it means that the form must always have `billable` set to something when it's initialized.

components/form/TaskForm.tsx
```diff
@@ -73,6 +75,20 @@ export function CreateTaskForm({ project, projectId, afterSubmit, onCancel }: Ta
                         }
                     />
                 </FormControl>
+                <FormControl py="0.5rem">
+                    <Flex flexDirection={'row'} alignItems={'baseline'}>
+                        <FormLabel>Billable</FormLabel>
+                        <Checkbox
+                            defaultIsChecked={taskFields.billable}
```
mapehe, line 82: The `defaultIsChecked` here is a mistake. The correct prop is `isChecked`.

mapehe (changes requested):

mapehe (approved):
Nice 💎

PR #110, opened 2022-01-28
feat: timesheets can be edited

oskarikerppo:


components/form/TimesheetForm.tsx
```diff
@@ -55,81 +55,151 @@ export function CreateTimesheetForm({
...
-            errorHandler(error as Error);
-        }
-    };
-
     const invalidAllocation =
         (timesheetFields.allocation && (timesheetFields.allocation < 0 || timesheetFields.allocation > 100)) || false;
 
+    const abortSubmission = onCancel ? onCancel : () => undefined;
+
+    const timesheetStatuses = { active: 'ACTIVE', inactive: 'INACTIVE', archived: 'ARCHIVED' };
```
mapehe, line 63: I'm not sure what's going on here, but I don't think you need these, since the openapi generator can infer these from the docs

https://github.com/three-consulting/epoc-frontend/blob/main/lib/types/api.ts#L129

mapehe (changes requested):
Nice job, this seems to work very well!

Some code looks unnecessary though, so might need to tidy up a little.

components/form/TimesheetForm.tsx
mapehe: Ah now I think I get why you wanted to do this. Maybe better to just not include this field in the form. The same way we handle the `project` field (Take a look at the `ProjectEditForm`).

oskarikerppo:
Screenshot from 2022-01-31 10-25-30

mapehe (approved):
LGTM

mapehe:
Please rebase to a single commit and feel free to merge 👏

PR #111, opened 2022-01-29
test: tests for create customer form

mapehe:
Introduce some basic tests that every form should pass and implement these for `CreateCustomerForm`:

* A form with only the required fields set is submitted correctly to the expected URL.
* A form with all fields set is submitted correctly to the expected URL.
* The method `afterSubmit` that [all forms have](https://github.com/three-consulting/epoc-frontend/blob/main/lib/types/forms.ts#L4) is invoked after submission as expected.
* The method `onCancel` that [all forms have](https://github.com/three-consulting/epoc-frontend/blob/main/lib/types/forms.ts#L5) is invoked on cancel as expected.

### DIFF
--- a/.github/workflows/housekeeping.yml
+++ b/.github/workflows/housekeeping.yml
@@ -26,3 +26,4 @@ jobs:
                   yarn run type-check
                   yarn prettier --check .
                   yarn eslint .
+                  yarn test
--- a/__tests__/component/form/CreateCustomerForm.test.tsx
+++ b/__tests__/component/form/CreateCustomerForm.test.tsx
@@ -1,0 +1,75 @@
+import React from 'react';
+import { render, screen, fireEvent, waitFor } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import { CreateCustomerForm } from '@/components/form/CustomerForm';
+import sinon, { spy } from 'sinon';
+import { Customer } from '@/lib/types/apiTypes';
+import { ApiUpdateResponse } from '@/lib/types/hooks';
+
+const customerEndpointURL = `${process.env.NEXT_PUBLIC_API_URL}/customer`;
+const bodySpy = sinon.spy((body) => body);
+const pathSpy = sinon.spy((path) => path);
+
+jest.mock('@/lib/utils/fetch', () => ({
+    post: async (path: string, body: object): Promise<ApiUpdateResponse<Customer>> => pathSpy(path) && bodySpy(body),
+}));
+
+const testCustomerRequiredFields: Customer = { name: 'Name' };
+const testCustomerAllFields: Customer = { name: 'anotherName', description: 'description' };
+
+afterEach(() => {
+    bodySpy.resetHistory();
+    pathSpy.resetHistory();
+});
+
+const fillAndSubmitForm = async (customer: Customer) => {
+    const nameInput = screen.getByTestId('form-field-name');
+    customer.name && fireEvent.change(nameInput, { target: { value: customer.name } });
+
+    const descriptionInput = screen.getByTestId('form-field-description');
+    customer.description && fireEvent.change(descriptionInput, { target: { value: customer.description } });
+
+    const submitButton = screen.getByTestId('form-button-submit');
+    await waitFor(() => fireEvent.click(submitButton));
+};
+
+test('a customer with the required fields only can be submitted', async () => {
+    render(<CreateCustomerForm />);
+    await fillAndSubmitForm(testCustomerRequiredFields);
+
+    await waitFor(() => expect(bodySpy.callCount).toEqual(1));
+    expect(pathSpy.getCalls()[0].args[0]).toStrictEqual(customerEndpointURL);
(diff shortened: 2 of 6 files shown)


mapehe:
Closes #92

__test__/component/form/CreateCustomerForm.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+afterEach(() => {
+    bodySpy.resetHistory();
+    pathSpy.resetHistory();
+});
+
+test('a customer with the required fields only can be submitted', async () => {
+    render(<CreateCustomerForm />);
+
+    const nameInput = screen.getByTestId('form-field-name');
+    fireEvent.change(nameInput, { target: { value: testCustomerRequiredFields.name } });
```
mapehe, line 29: Find name field by the data-testid attribute, that is set here:

https://github.com/three-consulting/epoc-frontend/blob/390347c6e069404141393b21ffca3792e1ce518c/components/form/CustomerForm.tsx#L43-L52

Then set that to the value `testCustomerRequiredFields.name`.

__test__/component/form/CreateCustomerForm.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+    const submitButton = screen.getByTestId('form-button-submit');
+    await waitFor(() => fireEvent.click(submitButton));
```
mapehe, line 32: Find submit button by `data-testid` attribute and click it.

__test__/component/form/CreateCustomerForm.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+    await waitFor(() => expect(bodySpy.callCount).toEqual(1));
+    expect(pathSpy.getCalls()[0].args[0]).toStrictEqual(customerEndpointURL);
+    expect(bodySpy.getCalls()[0].args[0]).toStrictEqual(testCustomerRequiredFields);
```
mapehe, line 36: Make sure that the mock-function defined here has been called:

https://github.com/three-consulting/epoc-frontend/blob/390347c6e069404141393b21ffca3792e1ce518c/__test__/component/form/CreateCustomerForm.test.tsx#L13-L15

__test__/component/form/CreateCustomerForm.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+import sinon, { spy } from 'sinon';
+import { Customer } from '@/lib/types/apiTypes';
+import { ApiUpdateResponse } from '@/lib/types/hooks';
+
+const customerEndpointURL = `${process.env.NEXT_PUBLIC_API_URL}/customer`;
+const bodySpy = sinon.spy((body) => body);
+const pathSpy = sinon.spy((path) => path);
+
+jest.mock('@/lib/utils/fetch', () => ({
+    post: async (path: string, body: object): Promise<ApiUpdateResponse<Customer>> => pathSpy(path) && bodySpy(body),
```
mapehe, line 14: Here we *mock*  the function `post` defined in `@/lib/utils/fetch`. What it means is that when anything attempts to call `post` from `@/lib/utils/fetch`, it calls this instead.

So what happens in this hook

https://github.com/three-consulting/epoc-frontend/blob/390347c6e069404141393b21ffca3792e1ce518c/lib/hooks/useCustomers.ts#L24-L28

is that the `post` in that hook is replaced by this mock function we define here.

The point is that, in this test, we only care about the fact that `post` has been called. Not about the actual thing that `post` does.

__test__/component/form/CreateCustomerForm.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { render, screen, fireEvent, waitFor } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import { CreateCustomerForm } from '@/components/form/CustomerForm';
+import sinon, { spy } from 'sinon';
+import { Customer } from '@/lib/types/apiTypes';
+import { ApiUpdateResponse } from '@/lib/types/hooks';
+
+const customerEndpointURL = `${process.env.NEXT_PUBLIC_API_URL}/customer`;
+const bodySpy = sinon.spy((body) => body);
+const pathSpy = sinon.spy((path) => path);
```
mapehe, line 11: Here we define a couple of "spies". Spy is essentially just a function that does some extra bookkeeping. What we care about here, is that after `bodySpy()` it's possible to tell that `bodySpy` has been invoked by taking a look at `bodySpy.getCalls()`.

__test__/component/form/CreateCustomerForm.test.tsx
```diff
@@ -0,0 +1,75 @@
+import React from 'react';
```
oskarikerppo, line 1: Shouldn't the test folder be named __tests__?

mapehe: Sure, let's call it tests

oskarikerppo (approved):

PR #120, opened 2022-01-31
test: test for create timesheet form

oskarikerppo:


yarn.lock
```diff
@@ -3970,6 +3970,13 @@ acorn-globals@^6.0.0:
     acorn "^7.1.1"
     acorn-walk "^7.1.1"
 
+acorn-jsx@^3.0.0:
```
mapehe, line 3973: I don't think we need any new deps and `yarn.lock` should be even with main.

__tests__/component/form/CreateTimesheetForm.test.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+jest.mock("@/lib/utils/fetch", () => ({
+    // eslint-disable-next-line require-await
+    post: async (path: string, body: object): Promise<ApiUpdateResponse<Timesheet>> => pathSpy(path) && bodySpy(body),
+}))
+
+// eslint-disable-next-line camelcase
+const testEmployee: Employee = { id: 1, first_name: "testi", last_name: "tekija", email: "test@test.test" }
+
+const testProject: Project = {
```
mapehe, line 25: You could create a file called `fixtures.ts` that exports these. The following fixtures from the `CreateCustomerForm` test could be moved to the same file:
```typescript
const testCustomerRequiredFields: Customer = { name: "Name" }
const testCustomerAllFields: Customer = { name: "anotherName", description: "description" }
```

mapehe (changes requested):
I suggest some tidying.

mapehe (approved):
Nice 💎, thanks for your patience

PR #121, opened 2022-01-31
style: very strict lint rules

mapehe:
Current eslint rules are quite lax. We should probably consider much stricter set of rules. I worked with these for a while and they would seem to enforce good code quality.

We can always turn some of these off later, if they are too annoying.


mapehe:
closes #107

oskarikerppo (approved):

PR #125, opened 2022-02-03
test: add tests for edit timesheet form

oskarikerppo:


__tests__/component/form/EditTimesheetForm.test.tsx
```diff
@@ -0,0 +1,208 @@
...
+        })
+    }
+
+    const employeeInput = screen.getByTestId("form-field-employee")
+    if (timesheet.employee) {
+        fireEvent.change(employeeInput, {
+            target: { value: timesheet.employee.id },
+        })
+    } else {
+        fireEvent.change(employeeInput, { target: { value: null } })
```
mapehe, line 65: Perhaps this else statement should be removed, since it's not possible to set `value` to `null` for this field.

__tests__/component/form/EditTimesheetForm.test.tsx
```diff
@@ -0,0 +1,208 @@
...
+    } else {
+        fireEvent.change(nameInput, { target: { value: "" } })
+    }
+
+    const descriptionInput = screen.getByTestId("form-field-description")
+    if (timesheet.description) {
+        fireEvent.change(descriptionInput, {
+            target: { value: timesheet.description },
+        })
+    }
```
mapehe, line 50: I think this should have a similar else statement to this

https://github.com/three-consulting/epoc-frontend/blob/07eba6d33d50cd476478145b33ec80982420e2c0/__tests__/component/form/EditTimesheetForm.test.tsx#L41-L43

__tests__/component/form/EditTimesheetForm.test.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+    const allocationInput = screen.getByTestId("form-field-allocation")
+    if (timesheet.allocation) {
+        fireEvent.change(allocationInput, {
+            target: { value: timesheet.allocation },
+        })
+    }
```
mapehe, line 57: I think this should have a similar else statement to this

https://github.com/three-consulting/epoc-frontend/blob/07eba6d33d50cd476478145b33ec80982420e2c0/__tests__/component/form/EditTimesheetForm.test.tsx#L41-L43

mapehe (changes requested):

__tests__/component/form/EditTimesheetForm.test.tsx
mapehe: Actually it's probably better to remove these if's and replace them with the one-liners like
```typescript
fireEvent.change(nameInput, { target: { value: timesheet.name || "" } })
```

__tests__/component/form/EditTimesheetForm.test.tsx
mapehe: Actually it's probably better to remove these if's and replace them with the one-liners like
```typescript
fireEvent.change(nameInput, { target: { value: timesheet.name || "" } })
```

mapehe (approved):
LGTM 👏

PR #127, opened 2022-02-07
Refactor/refactor task form

oskarikerppo:


components/form/TaskForm.tsx
```diff
@@ -36,6 +36,110 @@ const validateTaskFields = (fields: TaskFields, projectId: number): Task => {
...
+
+export function FormAlerts({ errorMessage }: FormAlertProps): JSX.Element {
+    return (
+        <>
+            <ErrorAlert />
+            <div>{errorMessage}</div>
+        </>
+    )
+}
+
```
mapehe, line 142: These should be exported in a separate file.

components/form/TaskForm.tsx
```diff
@@ -65,63 +169,25 @@ export function CreateTaskForm({
...
+                <BillableInput
+                    taskFields={taskFields}
+                    setTaskFields={setTaskFields}
+                />
+            </FormContainer>
+            <FormContainer>
+                <FromSubmitButton onClick={onSubmit} />
+                <FromCancelButton onClick={onCancel} />
+            </FormContainer>
+            {errorMessage && <FormAlerts errorMessage={errorMessage} />}
```
mapehe, line 190: The form looks pretty nice!

However, the components seem to be operating on a suboptimal level of abstraction. A component
```typescript
<DescriptionInput
      taskFields={taskFields}
      setTaskFields={setTaskFields}
/>
```
is not very recyclable, since it's only relevant if the form contains a field `description`.

I think the correct level of abstraction is something along the lines of
```typescript
<FormInputField
    label={”Description"}
    placeholder={"Description"}
    onChange={(e) => setTaskFields({...taskFields, description: e.target.value})}

... etc etc

 />
```
I.e. I think we should make a general "FormInputField" component that is defined once and can be used in all forms.

I'd make a separate component for the chechbox.

So in this form we should have two input fields and one checkbox field.

components/form/TaskForm.tsx
```diff
@@ -36,6 +36,110 @@ const validateTaskFields = (fields: TaskFields, projectId: number): Task => {
...
+            Submit
+        </Button>
+    )
+}
+
+export function FromCancelButton({ onClick }: FormButtonProps): JSX.Element {
+    return <Button onClick={onClick}>Cancel</Button>
+}
+
+export const FormContainer = styled.div`
```
mapehe, line 125: Nice, this is correct use of styled-components 👍

components/form/TaskForm.tsx
```diff
@@ -36,6 +36,110 @@ const validateTaskFields = (fields: TaskFields, projectId: number): Task => {
...
+            <Input
+                placeholder="Task Name"
+                onChange={(event) =>
+                    setTaskFields({
+                        ...taskFields,
+                        name: event.target.value,
+                    })
+                }
+            />
+            <FormErrorMessage>Task name cannot be empty.</FormErrorMessage>
```
mapehe, line 60: Should this be rendered conditionally?

mapehe (changes requested):
I think the component API should be redesigned to be more recyclable.

components/form/TaskForm.tsx
oskarikerppo: I think that's a chakra component that's only rendered if the isInvalid attribute equals true in the parent component

components/common/FormFields.tsx
```diff
@@ -0,0 +1,108 @@
...
+    Flex,
+    FormControl,
+    FormErrorMessage,
+    FormLabel,
+    Input,
+} from "@chakra-ui/react"
+import styled from "@emotion/styled"
+import ErrorAlert from "./ErrorAlert"
+
+type TaskFieldInputProps = {
```
mapehe, line 14: Should this be `FormInputFieldProps`?

components/common/FormFields.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+export function FormInputField({
+    label,
+    placeholder,
+    isRequired,
+    isInvalid,
+    formErrorMessage,
+    onChange,
+}: TaskFieldInputProps): JSX.Element {
+    if (isRequired) {
```
mapehe, line 31: Is this necessary? I think you should be able to render as follows:

```typescript
<FormControl isInvalid={!isInvalid} isRequired={isRequired}>
```

and if `isRequired` is false or undefined, the component works as intended.

components/common/FormFields.tsx
```diff
@@ -0,0 +1,108 @@
...
+                {formErrorMessage && (
+                    <FormErrorMessage>{formErrorMessage}</FormErrorMessage>
+                )}
+            </FormControl>
+        )
+    }
+    return (
+        <FormControl>
+            <FormLabel>{label}</FormLabel>
+            <Input placeholder={placeholder} onChange={onChange} />
```
mapehe, line 45: Shouldn't this have `isInvalid` set to something?

components/common/FormFields.tsx
```diff
@@ -0,0 +1,108 @@
...
+                <Checkbox isChecked={isChecked} onChange={onChange} />
+            </Flex>
+        </FormControl>
+    )
+}
+
+type FormButtonProps = {
+    onClick:
+        | ((() => void) | undefined)
+        | ((event: React.MouseEvent) => Promise<void | null>)
```
mapehe, line 77: This type looks wonky. Wouldn't something like

```typescript
onClick: (event: React.MouseEvent) => void
```

work?

components/form/TaskForm.tsx
```diff
@@ -65,63 +62,47 @@ export function CreateTaskForm({
...
+                        setTaskFields({
+                            ...taskFields,
+                            billable: event.target.checked,
+                        })
+                    }
+                />
+            </FormContainer>
+            <FormContainer>
+                <FromSubmitButton onClick={onSubmit} />
+                <FromCancelButton onClick={onCancel} />
```
mapehe, line 103: I think this could just be a single component
```typescript
<FormButtons onSubmit={onSubmit} onCancel={onCancel} />
```

components/common/FormFields.tsx
```diff
@@ -0,0 +1,108 @@
...
+            Submit
+        </Button>
+    )
+}
+
+export function FromCancelButton({ onClick }: FormButtonProps): JSX.Element {
+    return <Button onClick={onClick}>Cancel</Button>
+}
+
+export const FormContainer = styled.div`ChangeEventHandler<HTMLInputElement>
```
mapehe, line 92: What's the `ChangeEventHandler<HTMLInputElement>` doing here?

components/form/TaskForm.tsx
```diff
@@ -65,63 +62,47 @@ export function CreateTaskForm({
...
-                    Cancel
-                </Button>
-            </div>
-            {errorMessage && (
-                <>
-                    <ErrorAlert />
-                    <Box>{errorMessage}</Box>
-                </>
-            )}
+            <FormContainer>
```
mapehe, line 65: I'd prefer the containers to be arranged as 
```typescript
<FormContainer>
    <Stuff />
</FormContainer>
```

instead of
```typescript
<>
    <FormContainer>
        <Stuff />
    </FormContainer>
    <FormContainer>
        <Stuff />
    </FormContainer>
</>
```

If you want another container for the fields you can just define something like `FormFieldsContainer` separately and just nest it.

mapehe (changes requested):

components/common/FormFields.tsx
oskarikerppo: yes

components/common/FormFields.tsx
oskarikerppo: Screenshot from 2022-02-08 09-51-11

oskarikerppo: It will work, but the isInvalid will always think the description field is invalid. Because isRequired is false on the desciription field, the form will still go through but I don't think it is intended behavior

components/common/FormFields.tsx
oskarikerppo: It will work for onSubmit, yes

components/common/FormFields.tsx
oskarikerppo: that shouldn't be there

components/common/FormFields.tsx
```diff
@@ -0,0 +1,98 @@
...
+    Input,
+} from "@chakra-ui/react"
+import styled from "@emotion/styled"
+import ErrorAlert from "./ErrorAlert"
+
+type FormInputFieldProps = {
+    label: string
+    placeholder: string
+    isRequired: boolean
+    isInvalid?: string
```
mapehe, line 18: The correct type for `isInvalid` seems to be `boolean`.

components/common/FormFields.tsx
```diff
@@ -0,0 +1,98 @@
...
+export function FormInputField({
+    label,
+    placeholder,
+    isRequired,
+    isInvalid,
+    formErrorMessage,
+    onChange,
+}: FormInputFieldProps): JSX.Element {
+    return (
+        <FormControl isInvalid={!isInvalid} isRequired={isRequired}>
```
mapehe, line 32: We should definitely design the component API so that we set `isInvalid={isInvalid}` here.

components/form/TaskForm.tsx
```diff
@@ -65,63 +61,44 @@ export function CreateTaskForm({
...
+                <FormInputField
+                    label={"Name"}
+                    placeholder={"Task name"}
+                    isRequired={true}
+                    isInvalid={taskFields.name}
```
mapehe, line 69: I think the idiomatic way to set `isInvalid` here is `isInvalid={!taskFields.name}`.

mapehe (changes requested):

components/common/FormFields.tsx
```diff
@@ -0,0 +1,98 @@
...
+    FormLabel,
+    Input,
+} from "@chakra-ui/react"
+import styled from "@emotion/styled"
+import ErrorAlert from "./ErrorAlert"
+
+type FormInputFieldProps = {
+    label: string
+    placeholder: string
+    isRequired: boolean
```
mapehe, line 17: It would probably be simpler to set
```typescript
isRequired?: boolean
```
and simply leave it undefined if a component is not required.

components/form/TaskForm.tsx
```diff
@@ -65,63 +61,44 @@ export function CreateTaskForm({
...
+                        setTaskFields({
+                            ...taskFields,
+                            name: event.target.value,
+                        })
+                    }
+                />
+                <FormInputField
+                    label={"Description"}
+                    placeholder={"Description"}
+                    isRequired={false}
```
mapehe, line 81: After setting `isRequired` optional, we can drop this.

components/common/FormFields.tsx
mapehe: Yes this is not intended.  implementing my suggestions below should fix this.

components/common/FormFields.tsx
```diff
@@ -0,0 +1,98 @@
...
+type FormInputFieldProps = {
+    label: string
+    placeholder: string
+    isRequired?: boolean
+    isInvalid?: boolean
+    formErrorMessage?: string
+    onChange: ChangeEventHandler<HTMLInputElement>
+}
+
+export function FormInputField({
```
mapehe, line 23: This could be an arrow function
```typescript
export const FormInputField = ({
    label,
    placeholder,
    isRequired,
    isInvalid,
    formErrorMessage,
    onChange,
}: FormInputFieldProps): JSX.Element => (
    <FormControl isInvalid={isInvalid} isRequired={isRequired}>
        <FormLabel>{label}</FormLabel>
        <Input placeholder={placeholder} onChange={onChange} />
        {formErrorMessage && (
            <FormErrorMessage>{formErrorMessage}</FormErrorMessage>
        )}
    </FormControl>
)
```

components/common/FormFields.tsx
```diff
@@ -0,0 +1,98 @@
...
+    )
+}
+
+type CheckBoxFieldProps = {
+    label: string
+    isChecked: boolean
+    onChange: ChangeEventHandler<HTMLInputElement>
+}
+
+export function CheckBoxField({
```
mapehe, line 48: This could be an arrow function

components/common/FormFields.tsx
```diff
@@ -0,0 +1,98 @@
...
+        </FormControl>
+    )
+}
+
+type FormButtonProps = {
+    onSubmit: (event: React.MouseEvent) => void
+    onCancel: (() => void) | undefined
+}
+
+export function FromButtons({
```
mapehe, line 68: This could be an arrow function

components/common/FormFields.tsx
```diff
@@ -0,0 +1,98 @@
...
+export const FormContainer = styled.div`
+    padding: 20px;
+    text-align: right;
+`
+
+type FormAlertProps = {
+    errorMessage: string
+}
+
+export function FormAlerts({ errorMessage }: FormAlertProps): JSX.Element {
```
mapehe, line 91: This could be an arrow function

mapehe (changes requested):

mapehe (approved):
Nice job! 🙏

PR #130, opened 2022-02-08
test: added tests for create project form

oskarikerppo:


__tests__/fixtures.ts
```diff
@@ -7,10 +7,32 @@ export const testEmployee: Employee = {
...
+    endDate: "2022-02-09",
+    customer: testCustomer,
+    managingEmployee: testEmployee,
+}
+
+export const testProjectRequiredFields: Project = {
+    name: "Some project",
+    startDate: "2022-02-08",
+    customer: testCustomer,
+    managingEmployee: testEmployee,
```
mapehe, line 29: Perhaps we should add another employee fixture and set these testProjects to have different employees?

oskarikerppo: I went on and created another test customer and employee. Might as well have different choices for both of those.

mapehe (approved):

mapehe (changes requested):
Oops I'll approve after squash

mapehe (approved):

PR #131, opened 2022-02-09
feat: tasks can be edited

oskarikerppo:


components/form/TaskForm.tsx
```diff
@@ -98,9 +116,60 @@ export function CreateTaskForm({
...
+    }
+
+    return (
+        <>
+            <TaskForm {...props} onSubmit={onSubmit} />
+            {errorMessage && (
+                <>
+                    <ErrorAlert />
+                    <Box>{errorMessage}</Box>
+                </>
```
mapehe, line 171: We could probably use the `FormAlerts` component you recently built instead:

https://github.com/three-consulting/epoc-frontend/blob/c91049e326908e32543f2a272f6dce0b841ecc90/components/common/FormFields.tsx#L125-L130

components/form/TaskForm.tsx
```diff
@@ -98,9 +116,60 @@ export function CreateTaskForm({
...
```
mapehe, line 147: We could probably use the `FormAlerts` component you recently built instead:

https://github.com/three-consulting/epoc-frontend/blob/c91049e326908e32543f2a272f6dce0b841ecc90/components/common/FormFields.tsx#L125-L130

components/form/TaskForm.tsx
```diff
@@ -32,39 +46,42 @@ const validateTaskFields = (fields: TaskFields, projectId: number): Task => {
...
+            onSubmit(task)
         } catch (error) {
             errorHandler(error as Error)
-            return null
         }
     }
 
     return (
-        <>
+        <FormBackground>
```
mapehe, line 79: I'd prefer this to be called `FormContainer` and the current `FormContainer` to be called something like `FormFieldsContainer`.

Suggestion:

1. Navigate to the current `FormContainer` below on VSCode.
2. Press `f12` to goto definition.
3. Press `f2` to rename.
4. Navigate to the definition of `FormBackground`. Press `f2` to rename.

mapehe (changes requested):
Good work overall, some suggestions here and there.

mapehe (changes requested):
LGTM! Still need to resolve the conflicts and squash

mapehe:
Closing this as it's very old. (Sorry @oskarikerppo 😢 )

PR #132, opened 2022-02-09
feat: added customer table

oskarikerppo:
Screenshot from 2022-02-09 15-47-38

A Customer table is added below project table. The modal used to display create customer form is moved to FormFields.
Also the button "Add Customer" now uses a button from FormFields.


mapehe:
Good enough for me. I think we'll need to reorganize the UI at some point anyway. Feel free to squash.

mapehe (approved):

PR #147, opened 2022-05-05
fix: fix task post validation

mapehe:
As of `main`, there is a mismatch between SWR cache keys [here](https://github.com/three-consulting/epoc-frontend/blob/main/lib/hooks/useTasks.ts#L25-L26) and the corresponding mutate's [here](https://github.com/three-consulting/epoc-frontend/blob/main/lib/hooks/useTasks.ts#L39) and [here](https://github.com/three-consulting/epoc-frontend/blob/main/lib/hooks/useTasks.ts#L50). This breaks task post validation as pointed out by @robertsiipola.

It's probably easiest to just drop `taskByProjectEndpointURL` and use `taskEndpointURL` as a cache key throughout to amend this.


robertsiipola (approved):
👍🏼

ToukoH (approved):
LGTM

PR #152, opened 2022-05-09
feat: created customer detail page

ToukoH:


components/detail/CustomerDetail.tsx
```diff
@@ -0,0 +1,31 @@
+import { Customer } from "@/lib/types/apiTypes"
+import { Box, Flex, Heading } from "@chakra-ui/layout"
+import React from "react"
+
+type CustomerDetailprops = {
```
mapehe, line 5: You probably intended to write `CustomerDetailProps` here.

components/detail/CustomerDetail.tsx
```diff
@@ -0,0 +1,31 @@
...
+            border="1px solid"
+            borderColor="gray.400"
+            borderRadius="0.2rem"
+            padding="1rem 1rem"
+        >
+            <Heading>
+                <Box>{name}</Box>
+            </Heading>
+            <Flex>Description: {description}</Flex>
+            <Flex>Id: {id}</Flex>
```
mapehe, line 24: I think we should not display `id` here: It's a database row id for internal use only, so it's not really relevant information for the user.

components/detail/CustomerDetail.tsx
```diff
@@ -0,0 +1,31 @@
...
+            <Flex>Created: {created}</Flex>
```
mapehe, line 25: Currently these are ridiculously precise (like `2022-05-03T08:39:06.557445`). It would probably suffice to just convert these to dates.

lib/hooks/useCustomers.ts
```diff
@@ -12,11 +12,25 @@ import { NEXT_PUBLIC_API_URL } from "../conf"
 import { User } from "firebase/auth"
 
 export const customerEndpointURL = `${NEXT_PUBLIC_API_URL}/customer`
```
mapehe, line 14: For reasons I will explain later this file will probably need a `matchMutate`. Take a look at the `lib/hooks/useProjects.ts` file and look for match mutate.

You should define 
```
const projectIdEndpointCacheKeyRegex = /^\/project\/([0-9]+)$/
```
here and have the corresponding `matchMutate` for put and post hooks in this file below.

pages/customer/[id]/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+import Loading from "@/components/common/Loading"
+import CustomerDetail from "@/components/detail/CustomerDetail"
+import { UserContext } from "@/lib/contexts/FirebaseAuthContext"
+import { useCustomerDetail } from "@/lib/hooks/useCustomers"
+
+type Props = {
+    customerId: number
+}
+
+function CustomerDetailsPage({ customerId }: Props): JSX.Element {
```
Comment (mapehe) [inline pages/customer/[id]/index.tsx:15]: I'd prefer this to be called `CustomerDetailPage`.

pages/customer/[id]/index.tsx
```diff
@@ -0,0 +1,42 @@
+import React, { useContext } from "react"
```
Comment (mapehe) [inline pages/customer/[id]/index.tsx:1]: Could you add links to these detail pages on the list that the user finds on `/customer`. Clicking on a customer on the list should open the corresponding detail page.

mapehe (changes requested):
Nice work overall, some polishing and this will be very nice!

lib/types/api.ts
```diff
@@ -77,7 +77,7 @@ export interface components {
       name: string;
       description?: string;
       /** Format: date-time */
-      created?: DateConstructor;
+      created?: string;
```
ToukoH, line 80: Apparently I have managed to change the type of 'created' from string to DateConstructor due to some unknown reasons. I checked the master-branch and the type of 'created' was originally a string so I reverted it.

mapehe (changes requested):
Some very minor comments and we're all good!

components/detail/CustomerDetail.tsx
```diff
@@ -22,8 +22,8 @@ function CustomerDetail({ customer }: CustomerDetailProps): JSX.Element {
                 <Box>{name}</Box>
             </Heading>
             <Flex>Description: {description}</Flex>
-            <Flex>Created: {parseDate(created)}</Flex>
-            <Flex>Updated: {parseDate(updated)}</Flex>
+            <Flex>Created: {created && getDisplayDate(created)}</Flex>
```
mapehe, line 25: I think it makes more sense to render
```
created && <Flex>Created: {getDisplayDate(created)}</Flex>
```

mapehe (changes requested):

mapehe (commented):
Nice work! 💎  Let's squash this to a single commit and merge.

mapehe (approved):

PR #153, opened 2022-05-16
feat: customers can now be edited

ToukoH:


components/form/CustomerForm.tsx
```diff
@@ -18,80 +27,110 @@ const validateCustomerFields = (fields: CustomerFields): Customer => {
...
+        customerOrNull || {}
+    )
+
+    const [errorMessage, setErrorMessage] = useState<string>("")
+    const errorHandler = (error: Error) => setErrorMessage(`${error}`)
+
+    const abortSubmission = onCancel && onCancel
+
+    return (
+        <>
```
mapehe, line 45: The edit page might look a bit nicer if, instead of an empty fragment `<></>` the form was wrapped in the same kind of `Flex` that we have in `TimesheetForm.tsx`:

```typescript
<Flex
    flexDirection="column"
    backgroundColor="white"
    border="1px solid"
    borderColor="gray.400"
    borderRadius="0.2rem"
    padding="1rem 1rem"
>
```

mapehe, line 45: The edit page might look a bit nicer if, instead of an empty fragment `<></>` the form was wrapped in the same kind of `Flex` that we have in `TimesheetForm.tsx`:

```typescript
<Flex
    flexDirection="column"
    backgroundColor="white"
    border="1px solid"
    borderColor="gray.400"
    borderRadius="0.2rem"
    padding="1rem 1rem"
>
```

mapehe (commented):
Really nice work @ToukoH 💎👏 I found almost nothing to nag about! Just one weird import which we probably don't need and one small aesthetic thing. Feel free to squash this to single commit after having a look at my comments, and I'll mark this ✅ then!

PR #155, opened 2022-05-18
test: test for edit customer form

ToukoH:


__tests__/component/form/EditCustomerForm.test.tsx
```diff
@@ -0,0 +1,142 @@
...
+        const customerMissingRequired = _.omit(testCustomerAllFields, field)
+
+        /* eslint-disable no-await-in-loop */
+        await fillAndSubmitForm(customerMissingRequired as Customer)
+        await new Promise((resolve) =>
+            setTimeout(() => resolve(null), submitTimeout)
+        )
+        /* eslint-enable */
+
+        expect(pathSpy.callCount).toEqual(1)
```
mapehe, line 138: This is incorrect. This line asserts that the form HAS been sent, while it should assert that the form has NOT been sent (since a required field is missing). The correct line is
```
expect(pathSpy.callCount).toEqual(0)
```
The form and the tests should work so that this is true.

I'll take a look and get back to you about this.

mapehe (changes requested):

mapehe (approved):
LGTM!

PR #160, opened 2022-05-19
feat: homepage is now more readable

ToukoH:


components/editor/TimesheetEntryEditor.tsx
```diff
@@ -96,16 +97,43 @@ const DayEditor = ({
...
+                    <div key={`updateEntryEditor-${entry.id}`}>
+                        <Link
+                            key={`updateEntryEditor-${entry.id}`}
+                            onClick={() => setInEditState(!inEditState)}
+                            style={{
+                                marginRight: ".5rem",
+                                fontWeight: "bold",
+                            }}
+                        >
+                            Inspect
```
mapehe, line 133: I don't think `inspect` is very useful content for this `Link`. I think something like
```
7.5 hours ${project.name}
```
would be a more useful thing to display.

components/editor/TimesheetEntryEditor.tsx
```diff
@@ -38,6 +38,7 @@ const DayEditor = ({
 
     const displayEntries = entries.filter((entry) => entry.date === dateStr)
     const [timesheet, setTimesheet] = useState<Timesheet>()
+    const [inEditState, setInEditState] = useState<boolean>(false)
```
mapehe, line 41: This is "half-correct". While this does indeed work well for a single entry, it behaves in a little unexpected way when there are several entries: When you click on any of the entries **all** of the entries will enter edit-mode.

What you will need instead is to have this state **per entry**. The easiest way to achieve this is to put the stuff that's approximately between lines `86` and `139` to its own component and put this kind of line into that component. Then you should render such component for each `entry` and everything will work nicely.

components/editor/TimesheetEntryEditor.tsx
```diff
@@ -96,16 +97,43 @@ const DayEditor = ({
...
                             />
+                            <Link
+                                key={`updateEntryEditor-${entry.id}`}
+                                onClick={() => setInEditState(!inEditState)}
+                                style={{
+                                    marginRight: ".5rem",
+                                    fontWeight: "bold",
+                                }}
+                            >
+                                Close
```
mapehe, line 108: I think this button is unnecessary. This functionality could be attached to the `cancel` button of the form. I.e. the function on line 102 could be set as the `onCancel` of the `EditTimesheetEntryForm`.

mapehe (changes requested):
Thanks! This sort of works, but I have a few requests.

components/editor/TimesheetEntryEditor.tsx
```diff
@@ -24,6 +31,57 @@ const dateToString = (date: Date): string => {
...
 
+const TimesheetEntryRow = ({
+    entry,
+    deleteTimesheetEntry,
+    dateStr,
+    timeCategories,
+}: TimesheetEntryRowProps): JSX.Element => {
+    const { id } = entry
+    const projectId = entry.timesheet.project.id
+    const [inEditState, setInEditState] = useState<boolean>(false)
```
mapehe, line 42: I'd probably name these `edit` and `setEdit`.

components/editor/TimesheetEntryEditor.tsx
```diff
@@ -24,6 +31,57 @@ const dateToString = (date: Date): string => {
...
+
+    return (
+        <div key={`${entry.id}-container`}>
```
mapehe, line 45: This `key` prop can be removed.

components/editor/TimesheetEntryEditor.tsx
```diff
@@ -24,6 +31,57 @@ const dateToString = (date: Date): string => {
...
+            <Link
+                key={`updateEntryEditor-${entry.id}`}
```
mapehe, line 47: This `key` prop can be removed.

components/editor/TimesheetEntryEditor.tsx
```diff
@@ -24,6 +31,57 @@ const dateToString = (date: Date): string => {
...
+                onClick={() => setInEditState(!inEditState)}
+                style={{
+                    marginRight: ".5rem",
+                    fontWeight: "bold",
+                }}
+            >
+                {entry.quantity} Hours on {entry.timesheet.project.name}
```
mapehe, line 54: I'd suggest
```
{entry.quantity} hours on {entry.timesheet.project.name}
```
(no capital H)

components/editor/TimesheetEntryEditor.tsx
```diff
@@ -24,6 +31,57 @@ const dateToString = (date: Date): string => {
...
+            </Link>
+            {inEditState && projectId && id && (
+                <>
+                    {" "}
```
mapehe, line 58: This can be removed

components/editor/TimesheetEntryEditor.tsx
```diff
@@ -24,6 +31,57 @@ const dateToString = (date: Date): string => {
...
+                    <Link
+                        key={`delete-${entry.id}`}
+                        onClick={() => deleteTimesheetEntry(id)}
+                        style={{
+                            marginLeft: ".5rem",
+                            fontWeight: "bold",
+                        }}
+                    >
+                        Delete
+                    </Link>{" "}
```
mapehe, line 78: The `{" "}` can be removed.

components/editor/TimesheetEntryEditor.tsx
```diff
@@ -81,32 +139,21 @@ const DayEditor = ({
...
-                            </Link>
-                        </div>
-                    )
-                }
-                return null
-            })}
+            <ul>
+                {displayEntries.map((entry) => (
+                    <li
+                        key={`updateEntryEditor-${entry.id}`}
```
mapehe, line 145: I'd suggest setting
```typescript
key={`timesheet-entry-${entry.id}-editor-container`}
```

components/editor/TimesheetEntryEditor.tsx
```diff
@@ -24,6 +31,57 @@ const dateToString = (date: Date): string => {
...
+                    fontWeight: "bold",
+                }}
+            >
+                {entry.quantity} Hours on {entry.timesheet.project.name}
+            </Link>
+            {inEditState && projectId && id && (
+                <>
+                    {" "}
+                    <EditTimesheetEntryForm
+                        key={`updateEntryEditor-${entry.id}`}
```
mapehe, line 60: This `key` prop can be removed.

components/editor/TimesheetEntryEditor.tsx
```diff
@@ -24,6 +31,57 @@ const dateToString = (date: Date): string => {
...
+                        entry={entry}
+                        timesheet={entry.timesheet}
+                        projectId={projectId}
+                        date={dateStr}
+                        timeCategories={timeCategories}
+                        onCancel={() => setInEditState(!inEditState)}
+                        afterSubmit={() => setInEditState(!inEditState)}
+                    />
+                    <Link
+                        key={`delete-${entry.id}`}
```
mapehe, line 70: This `key` prop can be removed.

mapehe (commented):
Some nitty gritty, otherwise we're good!

PR #171, opened 2022-05-19
fix: drop mobile layout

mapehe:
We get several console errors from the current mobile layout. It also seems to be broken. I suggest that at this point we simply drop the mobile version and redo it after we decide how to design the frontend in general.


mapehe:
The erros are related to this https://github.com/chakra-ui/chakra-ui-docs/issues/138 and I don't feel like dealing with that kind of stuff atm.

ToukoH (approved):
👍

PR #172, opened 2022-05-19
feat: index page request states

mapehe:
The current index page doesn't handle the different XHR request states in any way. For example, when there is an error, it simply displays nothing. Amend this situation and render by XHR-request state.


ToukoH (approved):
LGTM

PR #177, opened 2022-05-24
feat: created a page for reports

ToukoH:


mapehe (changes requested):
Two suggestions:

* I think this the link to the sidebar could be added in this PR.
* I'd prefer the URL to be `/report` instead of `/reports` (for consistency with the other URLs).

mapehe (commented):
LGTM! Let's squash to a single commit and I'll ✅ then

mapehe (approved):
Nice!

PR #179, opened 2022-05-30
feat: hook for fetching timesheet entries for a given time interval

ToukoH:


lib/hooks/useTimesheetEntries.ts
```diff
@@ -51,6 +51,24 @@ export const useEmployeeTimesheetEntries = (
     return { isSuccess: false, isError: false, isLoading: true }
 }
 
+export const useTimeIntervalTimesheetEntries = (
+    user: User
+): ApiGetResponse<TimesheetEntry[]> => {
+    const startDate = "2022-01-01"
+    const endDate = "2022-12-12"
+    if (startDate && endDate !== null) {
```
mapehe, line 59: This conditional statement is redundant since `startDate` and `endDate` are always defined.

lib/hooks/useTimesheetEntries.ts
```diff
@@ -51,6 +51,24 @@ export const useEmployeeTimesheetEntries = (
     return { isSuccess: false, isError: false, isLoading: true }
 }
 
+export const useTimeIntervalTimesheetEntries = (
```
mapehe, line 54: I think we could call this `useTimesheetEntries`.

pages/report/index.tsx
```diff
@@ -1,13 +1,20 @@
...
-const Report: NextPage = () => (
-    <div>
-        <Heading fontWeight="black" margin="1rem 0rem">
-            Reports
-        </Heading>
-    </div>
-)
+const Report: NextPage = () => {
+    const { user } = useContext(UserContext)
+    const reportsRespose = useTimeIntervalTimesheetEntries(user)
```
mapehe, line 9: We could get rid of the linter error here by replacing the current return value with
```typescript
    return (
        <div>
            {reportsRespose.isSuccess && <Heading fontWeight="black" margin="1rem 0rem">
                Reports
            </Heading>}
        </div>
    )
```

mapehe (changes requested):
Nice, but some polishing needed

mapehe (approved):
💎

PR #183, opened 2022-06-02
refactor: refactor hooks

mapehe:
Things are getting messy enough, that we need a generic way to interface with SWR. While keeping things simple was a good idea at first, the hooks are getting messy enough that it's difficult to manually avoid things like cache-key collisions and its also hard to remember which data needs to be depricated when a POST or PUT is performed. Most of the action is [here](https://github.com/three-consulting/epoc-frontend/pull/183/files#diff-90a5a44810a52267ace6948883c4ac5526ec21b1779dd86121f709cd2701e478).

I've also performed some related refactoring. Previously hooks were arranged by type (`Customer`, `Project` etc.). I think, given our new generics, it's better to arrange by functionality (`getDetail`, `getList`, `update`).

Here I've refactored our hooks. We make heavier use of generics than before and are hence able to drop a lot of repetitive error-prone code.


ToukoH (approved):
The functionality of the page remained the same. LGTM

PR #188, opened 2022-06-06
fix: post validation endpoint regex

mapehe:
The regex `^/${endpoint}[/|?].+` does match, `/customer/2`, `/customer?asd=asd` etc. but it does not match `/customer` (without the backslash at the end). This breaks the post validation when a new customer is created because the customer list is fetched from `/customer`.

I think the suffix `[/|?].+` should be optional, so the correct regex should be `^/${endpoint}([/|?].+)?`.


ToukoH (approved):
LGTM

PR #197, opened 2022-06-12
feat: filter report details by employee

mapehe:
Filter report by employee.


ToukoH (changes requested):
I don't know if this functionality is desired or not but reports page won't display the projects, customers or tasks unless the user has at least one timesheet entry. For example, I ran the shell script you wrote which removed all entries for me. After that I added myself into various different projects, but the page did not display any information until I created a single timesheet entry.

ToukoH (approved):
LGTM

PR #202, opened 2022-06-15
feat: the projected income is now displayed

ToukoH:


components/table/ReportTable.tsx
```diff
@@ -58,6 +59,13 @@ interface ReportTableProps {
 const entriesQuantitySum = (entries: TimesheetEntry[]) =>
     entries.reduce((total, currentItem) => total + currentItem.quantity, 0)
 
+const totalRate = (entries: TimesheetEntry[]) =>
```
mapehe, line 62: I think `totalIncome` would be more apt. I think rate should be measured in `€/h` while income is in `€`.

components/table/ReportTable.tsx
```diff
@@ -94,16 +102,18 @@ const entriesByTask = (entries: TimesheetEntry[], taskId: number) =>
 const taskByProject = (tasks: Task[], projectId: number) =>
     tasks.filter((task) => task.project.id === projectId)
 
-function TotalHours({
+function TotalHoursAndRate({
```
mapehe, line 105: Should we simply call this `Total`?

mapehe (commented):
Some tiny ideas.

PR #203, opened 2022-06-17
feat: date picker

ToukoH:
Screenshot 2022-06-17 at 13 26 12

As far as I know, the functionality of the date picker is correct, but I'm uncertain how the error handling should be done. Now when the input is invalid a new error is thrown, but it's not displayed in any proper way so if you'd like a box or some other component to be shown just let me know.

Screenshot 2022-06-17 at 13 32 16

### DIFF
--- a/components/table/ReportTable.tsx
+++ b/components/table/ReportTable.tsx
@@ -8,7 +8,23 @@ import {
     Timesheet,
     TimesheetEntry,
 } from "@/lib/types/apiTypes"
-import { Select, Checkbox } from "@chakra-ui/react"
+import {
+    Select,
+    Checkbox,
+    InputGroup,
+    Input,
+    Button,
+    FormControl,
+    FormErrorMessage,
+    FormHelperText,
+} from "@chakra-ui/react"
+
+interface DateInputProps {
+    startDate: string
+    endDate: string
+    setStartDate: React.Dispatch<string>
+    setEndDate: React.Dispatch<string>
+}
 
 interface TotalHoursProps {
     startDate: string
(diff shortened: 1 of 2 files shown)


components/table/ReportTable.tsx
```diff
@@ -102,6 +111,57 @@ const entriesByTask = (entries: TimesheetEntry[], taskId: number) =>
...
+                <Input
+                    placeholder="Start date (YYYY-MM-DD)"
+                    value={newStartDate}
+                    onChange={handleStartDateChange}
+                ></Input>
+                <Input
+                    placeholder="End date (YYYY-MM-DD)"
+                    value={newEndDate}
+                    onChange={handleEndDateChange}
+                ></Input>
```
LeadingMoominExpert, line 151: The inputs should be of type date so you don't need all the regex validation for the date being valid. 
See the form for creating a new project for example https://github.com/three-consulting/epoc-frontend/blob/main/components/form/ProjectForm.tsx#L147-L179
Can be found at url `/project/new`

LeadingMoominExpert (changes requested):

components/table/ReportTable.tsx
robertsiipola: @ToukoH you could still validate that the start date is before the end date, when both are chosen. You can use the `FormErrorMessage` component to display the error message: https://chakra-ui.com/docs/components/form/form-control#error-message

LeadingMoominExpert (approved):
Ok. I do suggest that you put more thought into the commit messages already when first making them. Half of the commit messages being `asd` and the rest all having the same name make the pr impossible to read per commit. Commit only proper messages 👍🏻

robertsiipola:
> Ok. I do suggest that you put more thought into the commit messages already when first making them. Half of the commit messages being `asd` and the rest all having the same name make the pr impossible to read per commit. Commit only proper messages 👍🏻

Additionally I think re-using commit messages in the same pull request is hazardous. Leaving aside the `asd` ones, each `feat: date-picker` commit should be more descriptive imo.

PR #216, opened 2022-08-04
fix: hide all but sign in when user is logged out

miikahyttinen:
- Control user state in Main-component
- Hide all but sign in link if user is not logged in

### DIFF
--- a/components/common/Main.tsx
+++ b/components/common/Main.tsx
@@ -1,26 +1,41 @@
 import React, { ReactNode } from "react"
 import { Box, Flex } from "@chakra-ui/react"
 import LeftNav from "./LeftNav"
+import useFirebaseAuth from "@/lib/hooks/useFirebaseAuth"
 
 interface MainProps {
     children?: ReactNode
 }
 
-const Main = ({ children }: MainProps): JSX.Element => (
-    <Box
-        display="block"
-        flex="1 0 auto"
-        marginLeft="5rem"
-        marginRight="auto"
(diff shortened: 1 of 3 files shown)


components/common/Main.tsx
```diff
@@ -1,26 +1,41 @@
-import React, { ReactNode } from "react"
+import React, { ReactNode, useContext } from "react"
```
LeadingMoominExpert, line 1: Unused import `useContext`

LeadingMoominExpert (approved):
lgtm, just one unused import that can be removed

lib/contexts/FirebaseAuthContext.tsx
```diff
@@ -2,28 +2,26 @@ import React, { createContext, ReactNode } from "react"
...
 export const AuthProvider = ({ children }: AuthProps): JSX.Element => {
     const auth = useFirebaseAuth()
-
     return <AuthContext.Provider value={auth}>{children}</AuthContext.Provider>
 }
 
 export const UserProvider = ({ children }: AuthProps): JSX.Element => {
     const { user } = useFirebaseAuth()
-
     return user ? (
```
mapehe, line 22: How about
```typescript
    return user && (
        <UserContext.Provider value={{ user }}>{children}</UserContext.Provider>
    ) 
```
instead of the ternary.

mapehe (commented):
One small nit. Otherwise LGTM

PR #220, opened 2022-08-09
feat: autosync required fields

mapehe:


script/generateTypeMetadata.js
```diff
@@ -0,0 +1,37 @@
...
+const readline = require("readline")
+
+// eslint-disable-next-line id-length
+const rl = readline.createInterface({
+    input: process.stdin,
+    output: process.stdout,
+    terminal: false,
+})
+
+const fieldToMetadata = (field, required) =>
```
mapehe, line 11: Maybe `generateFieldMetadata` would describe this thing better.

script/generateTypeMetadata.js
```diff
@@ -0,0 +1,37 @@
...
+    `${field}: {required: ${required.includes(field)}}`
+
+const schemaToString = (type, schema) => {
```
mapehe, line 14: Maybe `generateSchemaMetadata` would describe this thing better.

script/generateTypeMetadata.js
```diff
@@ -0,0 +1,37 @@
...
+    const fields = Object.keys(schema.properties)
+    const { required } = schema
+    const fieldMetadata = `export const ${type.toLowerCase()}FieldMetadata: Record<string, FieldMetadata> = {${fields
+        .map((field) => `${fieldToMetadata(field, required)}`)
+        .join(", ")}}`
+
+    return `${fieldMetadata}`
```
mapehe, line 21: This should probably be dropped

script/generateTypeMetadata.js
```diff
@@ -0,0 +1,37 @@
...
+    terminal: false,
+})
+
+const fieldToMetadata = (field, required) =>
+    `${field}: {required: ${required.includes(field)}}`
+
+const schemaToString = (type, schema) => {
+    const fields = Object.keys(schema.properties)
+    const { required } = schema
+    const fieldMetadata = `export const ${type.toLowerCase()}FieldMetadata: Record<string, FieldMetadata> = {${fields
```
mapehe, line 17: Instead of defining this const `fieldMetadata`, we  could just return the right hand side of the assignment.

script/generateTypeMetadata.js
```diff
@@ -0,0 +1,37 @@
...
+        .map((field) => `${fieldToMetadata(field, required)}`)
+        .join(", ")}}`
+
+    return `${fieldMetadata}`
+}
+
+rl.on("line", (data) => {
+    const schema = JSON.parse(data)
+    const types = schema.components.schemas
+    const typesStr = Object.keys(types)
```
mapehe, line 27: Perhaps rename this to `schemaMetadata`

script/generateTypeMetadata.js
mapehe, line 17: `type.toLowerCase()` is not the best idea here. For example `TimesheetEntry` becomes `timesheetentryFieldMetadata` when it should be `timesheetEntryFieldMetadata`.

script/generateTypeMetadata.js
```diff
@@ -0,0 +1,40 @@
...
+})
+
+const generateFieldMetadata = (field, required) =>
+    `${field}: {required: ${required.includes(field)}}`
+
+const generateSchemaMetadata = (type, schema) => {
+    const fields = Object.keys(schema.properties)
+    const { required } = schema
+    return `export const ${
+        type[0].toLowerCase() + type.slice(1)
```
mapehe, line 18: This could be a small separate utility function

PR #226, opened 2022-08-17
Updated types to match with backend

ronikeuru:
- Changed types script to use heroku as source
- Updated types
- Fixed types of employees in tests

### DIFF
--- a/__tests__/fixtures.ts
+++ b/__tests__/fixtures.ts
@@ -11,6 +11,8 @@ export const testEmployee: Employee = {
     firstName: "testi",
     lastName: "tekija",
     email: "test@test.test",
+    role: "USER",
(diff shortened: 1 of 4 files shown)


__tests__/fixtures.ts
```diff
@@ -11,13 +11,15 @@ export const testEmployee: Employee = {
     firstName: "testi",
     lastName: "tekija",
     email: "test@test.test",
+    role: "USER",
```
LeadingMoominExpert, line 14: Even though uid is not a mandatory field you should include it in the tests, because hopefully we'll end up in a situation once the initial db syncing is done that we can make it a mandatory field

LeadingMoominExpert (changes requested):

PR #227, opened 2022-08-17
Updated Github actions for type validation and added actions for pull request

ronikeuru:
- Added Github actions for pull request
- Updated type validation scripts
- Updated main and housekeeping actions with correct type validation scripts

### DIFF
--- a/.github/workflows/housekeeping.yaml
+++ b/.github/workflows/housekeeping.yaml
@@ -25,4 +25,4 @@ jobs:
           yarn prettier --check .
           yarn eslint .
           yarn test
-          yarn validate-types
+          yarn validate-types-dev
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yaml
```diff
@@ -23,6 +23,8 @@ jobs:
           yarn prettier --check .
           yarn eslint .
           yarn test
+          yarn validate-types-dev
+          yarn validate-types-prod
```
LeadingMoominExpert, line 27: These probably shouldn't be in the main flow, because we don't want deployment to the dev environment to fail if the production types don't match

LeadingMoominExpert (changes requested):

LeadingMoominExpert (approved):

PR #233, opened 2022-08-23
feat: hide navigation links from non admin users

miikahyttinen:
- hide projects, customers, and reports links from the left nav if user is not an admin

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -35,6 +35,8 @@ module.exports = {
         "id-blacklist": 0,
         "operator-linebreak": 0,
         "no-mixed-operators": 0,
+        "no-shadow": "off",
+        "@typescript-eslint/no-shadow": ["error"],
     },
     settings: {
         react: {
--- a/components/common/LeftNav.tsx
+++ b/components/common/LeftNav.tsx
@@ -1,4 +1,4 @@
-import React from "react"
+import React, { useContext } from "react"
 import { Text, Flex, Icon, HStack } from "@chakra-ui/react"
 import {
     BsBriefcase,
@@ -10,7 +10,8 @@ import {
 } from "react-icons/bs"
 import Link from "next/link"
 import { IconType } from "react-icons"
-import { User } from "firebase/auth"
+import { Role } from "@/lib/types/auth"
+import { AuthContext } from "@/lib/contexts/FirebaseAuthContext"
 
 interface LinkItemProps {
     name: string
@@ -18,14 +19,9 @@ interface LinkItemProps {
     href: string
 }
 
-interface LeftNavProps {
-    user: User | null
-    signInWithGoogle: () => Promise<void>
-    signOutAndClear: () => Promise<void>
-}
+const userLinkItem: LinkItemProps = { name: "Home", icon: BsHouse, href: "/" }
 
-const LinkItems: Array<LinkItemProps> = [
-    { name: "Home", icon: BsHouse, href: "/" },
+const adminLinkItems: LinkItemProps[] = [
     { name: "Projects", icon: BsBriefcase, href: "/project" },
     { name: "Customers", icon: BsPersonBadge, href: "/customer" },
     { name: "Reports", icon: BsFillFileTextFill, href: "/report" },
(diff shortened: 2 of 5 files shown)


.eslintrc.js
```diff
@@ -35,6 +36,7 @@ module.exports = {
         "id-blacklist": 0,
         "operator-linebreak": 0,
         "no-mixed-operators": 0,
+        "no-shadow": "off",
```
miikahyttinen, line 39: bug with typescript: https://stackoverflow.com/questions/63961803/eslint-says-all-enums-in-typescript-app-are-already-declared-in-the-upper-scope

components/common/LeftNav.tsx
```diff
@@ -24,8 +24,14 @@ interface LeftNavProps {
     signOutAndClear: () => Promise<void>
 }
 
-const LinkItems: Array<LinkItemProps> = [
-    { name: "Home", icon: BsHouse, href: "/" },
+enum Role {
```
mapehe, line 27: I'd prefer
```typescript
type Role = 'ADMIN' | 'USER'
```
We could then also revert the `.eslintrc.json` change, because I'm not completely sure about it's implications.

components/common/LeftNav.tsx
```diff
@@ -55,6 +61,21 @@ function LeftNav({
...
+        setRole(undefined)
+    }
+
+    useEffect(() => {
+        user?.getIdTokenResult(false).then((data) =>
+            data.claims.role
+                ? setRole(data.claims.role as Role)
+                : setRole(undefined)
+        )
+    }, [user])
```
mapehe, line 77: While this probably works, I'd prefer avoiding side effects. We could probably move this one component upwards and pass `role` of type `Role` as a parameter to this component. This would also allow us to drop the `useState` on line 64.

mapehe: Btw, I think we could strive to maximise idiomacy by expressing through types that `User` and `Role` must always exist together: If one does not exist, the other should not exist either. I'd suggest the following.
```typescript
interface UserAndRole {
    user: User
    role: Role
}

interface LeftNavProps {
    userAndRole: UserAndRole | null
    signInWithGoogle: () => Promise<void>
    signOutAndClear: () => Promise<void>
}
```
this could also make the render statement slightly more readible.

components/common/LeftNav.tsx
```diff
@@ -55,6 +61,21 @@ function LeftNav({
     signInWithGoogle,
     signOutAndClear,
 }: LeftNavProps): JSX.Element {
+    const [role, setRole] = useState<Role | undefined>(undefined)
+
+    const handleSignOut = () => {
```
mapehe, line 66: If we go a component upwards as described in the component below, we can remove this, no?

.eslintrc.js
mapehe: Nice, but I'd hesitate to make this change if we don't need it right now (see below), since bumping a version later on should suffice.

mapehe (changes requested):
Some suggestions

components/common/LeftNav.tsx
miikahyttinen: I would like to use an enum here. Then I can use it as a constant when checking role e.g. `role == Role.ADMIN`. When it comes to `.eslintrc.json`, I have replaced eslint's `no-shadow` rule with TypeScript specific rule, which seems to be a correct way: https://typescript-eslint.io/rules/no-shadow/ 

Otherwise, linter will give an error always when an enum is used. Fore sure, we can choose not to use enums at all 🙂

.eslintrc.js
miikahyttinen: https://typescript-eslint.io/rules/no-shadow/

components/common/LeftNav.tsx
mapehe: Fair enough, let's use enum. 🙏

components/common/Main.tsx
```diff
@@ -26,7 +37,7 @@ const Main = ({ children }: MainProps): JSX.Element => {
                 justifyContent="center"
             >
                 <LeftNav
-                    user={user}
+                    userWithRole={{ user, role }}
```
mapehe, line 40: Just a side note: Should we actually store `role` in `FirebaseAuthContext` instead of a locally scoped `const`?

miikahyttinen: It possible to refactor to be in context in future.

PR #256, opened 2022-09-20
feat: reactivate project

miikahyttinen:


package.json
```diff
@@ -8,6 +8,7 @@
     "scripts": {
         "dev": "next dev",
         "dev-local": "export JWT_ISSUER=$(gopass show --password 3/epoc/firebase/dev/SPRING_SECURITY_OAUTH2_RESOURCESERVER_JWT_ISSUER-URI) && export NEXT_PUBLIC_FIREBASE_API_KEY=$(gopass 3/epoc/firebase/dev/NEXT_PUBLIC_FIREBASE_API_KEY) && export NEXT_PUBLIC_FIREBASE_AUTH_DOMAIN=$(gopass 3/epoc/firebase/dev/NEXT_PUBLIC_FIREBASE_AUTH_DOMAIN) && export FIREBASE_SERVICE_ACCOUNT_JSON=$(gopass 3/epoc/firebase/dev/sa) && export NEXT_PUBLIC_API_URL=http://localhost:8080 && concurrently yarn:heroku-login yarn:start-server yarn:dev",
+        "dev-local-backend": "export JWT_ISSUER=$(gopass show --password 3/epoc/firebase/dev/SPRING_SECURITY_OAUTH2_RESOURCESERVER_JWT_ISSUER-URI) && export NEXT_PUBLIC_FIREBASE_API_KEY=$(gopass 3/epoc/firebase/dev/NEXT_PUBLIC_FIREBASE_API_KEY) && export NEXT_PUBLIC_FIREBASE_AUTH_DOMAIN=$(gopass 3/epoc/firebase/dev/NEXT_PUBLIC_FIREBASE_AUTH_DOMAIN) && export FIREBASE_SERVICE_ACCOUNT_JSON=$(gopass 3/epoc/firebase/dev/sa) && export NEXT_PUBLIC_API_URL=http://localhost:8080 && concurrently yarn:dev",
```
miikahyttinen, line 11: I created this new script because at least I need to develop frontend against local backend from time to time.

LeadingMoominExpert: I don't get what this is doing? Are you using this when already running database in docker and the backend from idea? If so, you can drop the envs `JWT_ISSUER` and `FIREBASE_SERVICE_ACCOUNT_JSON`. Perhaps the concurrently call is also redundant as you're just running one yarn command

miikahyttinen: Yes exactly. I dropped those env variables but I suppose `concurrently` is needed.

PR #264, opened 2022-10-11
update dev script name

miikahyttinen:
A proposal to rename dev scripts :) TODO:  fix gh actions if this propsal proceeds


ronikeuru (approved):

LeadingMoominExpert:
You could also introduce a new run script that points to the dev backend in heroku, so there's no need to always run everything locally

ronikeuru (approved):

PR #271, opened 2022-11-22
feat: admin can inspect and edit entries by other users

ToukoH:


LeadingMoominExpert (approved):
gg. These pages seem to be in area uncovered by the tests, which should be addressed as well

PR #273, opened 2022-11-24
Handle unauthorized errors with auth error alert

ronikeuru:
- Added new error alert component for unauthorized errors
  - The alert will be similar in all pages and therefore own component is useful
- Implemented the new component in pages of:
  - Employee
  - Report

### DIFF
--- a/components/table/ReportTable.tsx
+++ b/components/table/ReportTable.tsx
@@ -21,12 +21,14 @@ import {
 import { dateTimeToShortISODate, toLocalDisplayDate } from "@/lib/utils/date"
 import { round } from "lodash"
 import { NEXT_PUBLIC_API_URL } from "@/lib/conf"
-import { UserContext } from "@/lib/contexts/FirebaseAuthContext"
+import { AuthContext, UserContext } from "@/lib/contexts/FirebaseAuthContext"
 import { getText } from "@/lib/utils/fetch"
 import { downloadFile } from "@/lib/utils/common"
 import { useTimesheetEntries } from "@/lib/hooks/useList"
 import { DateTime } from "luxon"
 import ErrorAlert from "../common/ErrorAlert"
+import { Role } from "@/lib/types/auth"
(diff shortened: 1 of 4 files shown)


components/table/ReportTable.tsx
```diff
@@ -359,15 +361,20 @@ function ReportTable({
...
     const [endDate, setEndDate] = useState<string>(lastDay)
 
-    const reportsResponse = useTimesheetEntries(user, startDate, endDate)
-    const allEntries = reportsResponse.isSuccess ? reportsResponse.data : []
+    const reportsResponse =
+        role === Role.ADMIN
+            ? useTimesheetEntries(user, startDate, endDate)
+            : false
+    const allEntries =
+        reportsResponse && reportsResponse.isSuccess ? reportsResponse.data : []
```
LeadingMoominExpert, line 377: Type of `reportsResponse` is a little confusing here. Maybe it could be undefined if the role isn't admin? that's easy to verify in the next ternary too.

LeadingMoominExpert (changes requested):

components/table/ReportTable.tsx
```diff
@@ -372,9 +372,8 @@ function ReportTable({
     const reportsResponse =
         role === Role.ADMIN
             ? useTimesheetEntries(user, startDate, endDate)
-            : false
-    const allEntries =
-        reportsResponse && reportsResponse.isSuccess ? reportsResponse.data : []
+            : undefined
+    const allEntries = reportsResponse?.isSuccess ? reportsResponse.data : []
```
LeadingMoominExpert, line 376: `!!reportsResponse` might work here as well but it's up to you

LeadingMoominExpert (approved):

components/table/ReportTable.tsx
ronikeuru: Field `reportsResponse.data` exists iff `reportsResponse.isSuccess === true`

PR #283, opened 2023-02-08
278 improve ux of selecting multiple dates for inputting workhours

ronikeuru:
- Improved the date multi select UX
- Improved the UI look
- Turned parts of the app into reusable components


components/common/Buttons.tsx
```diff
@@ -0,0 +1,125 @@
...
+export const AddButton = (props: IButton): JSX.Element => (
+    <StyledButton colorScheme="blue" variant="solid" {...props}>
+        {`Add${props.name ? " ".concat(props.name) : ""}`}
+    </StyledButton>
+)
+
+export const EditButton = (props: IButton): JSX.Element => (
+    <StyledButton colorScheme="blue" variant="solid" {...props}>
+        {`Edit${props.name ? " ".concat(props.name) : ""}`}
+    </StyledButton>
```
LeadingMoominExpert, line 59: Lots of repetitive code with the separate buttons. Could you make a unexported function teking color and the name as inputs and return the StyledButton component and use that for all of these?

components/common/Header.tsx
```diff
@@ -1,19 +1,104 @@
...
+        </Heading>
+    </StyledItems>
+)
+
+export const SubHeader = ({ text }: { text: string }) => (
+    <StyledItems>
+        <Heading as="h3" variant="subHeader" size="md">
+            {text}
+        </Heading>
+    </StyledItems>
```
LeadingMoominExpert, line 70: These headers also contain a lot of repetitive code which could be extracted to one function taking the `as`, `variant`, `size` and the contents as parameters

components/form/EmployeeForm.tsx
```diff
@@ -67,6 +68,42 @@ const EmployeeForm = ({ onSubmit, onCancel, employee }: EmployeeFormProps) => {
         }
     }, [employeeFields])
 
+    const onConfirm = (event: React.MouseEvent) => {
+        event.preventDefault()
+        try {
+            onSubmit(validateEmployeeFields(employeeFields))
+        } catch (error) {
+            errorHandler(error as Error)
```
LeadingMoominExpert, line 76: Not really an issue of this pr but is the casting really needed if the errorHandler already accepts that as type? What happens without the type cast?

components/common/FormSection.tsx
```diff
@@ -0,0 +1,38 @@
...
+    children: JSX.Element | Array<JSX.Element>
+}
+
+const FormSection = (props: IFormSectionProps): JSX.Element => {
+    const { header, children, errorMessage } = props
+    return (
+        <>
+            <Box paddingX="2rem" paddingY="1rem">
+                <SubHeader text={header} />
+                <Flex
```
mapehe, line 18: Perhaps related to @LeadingMoominExpert comments, I would prefer us to find a way to separate styles from render statements such as this one right here. I really like the way things were done in my previous project, [see e.g. this](https://github.com/espoon-voltti/evaka/blob/master/frontend/src/employee-frontend/components/finance-basics/FeeThresholdsItem.tsx#L205-L222). I think something that looks like that would promote recycling styles and avoid repetition.

Might be too much work to tackle in this PR though.

components/form/EmployeeForm.tsx
ronikeuru: epoc_errorhandling_unknown

components/common/Header.tsx
```diff
@@ -46,37 +48,41 @@ const StyledItems = ({ children }: { children: JSX.Element }) => (
...
-)
+const Header = ({ children, type }: IHeader) => {
+    let [variant, size]: [THeaderType, string] = ["top", "3xl"]
+    switch (type) {
+        case "main": {
+            ;[variant, size] = [type, "xl"]
+            break
+        }
+        case "sub": {
+            ;[variant, size] = [type, "md"]
```
LeadingMoominExpert, line 67: These unnecessary semicolons are breaking the checks

LeadingMoominExpert (approved):
Good stuff! Just fix the linter issues

components/common/FormSection.tsx
ronikeuru: Not in scope of this PR

PR #301, opened 2023-05-03
295 v2 create alternative styles for mobile devices

ronikeuru:
- Updated yarn packages
- Added alternative styles for mobile devices


LeadingMoominExpert:
I think the components should scale with the width of the screen not width of the content text
image

LeadingMoominExpert (approved):

PR #310, opened 2023-05-11
Filter archived employees from form selections

venlavanhalathree:
closes#308


components/form/TimesheetForm.tsx
```diff
@@ -123,6 +123,8 @@ function TimesheetForm({
 
     const abortSubmission = onCancel && onCancel
 
+    const actives = employees.filter((employee) => employee.status === "ACTIVE")
```
LeadingMoominExpert, line 126: Same here, use `activeEmployees` to be more descriptive

PR #317, opened 2023-05-19
Limit the working hours on entry form

### LINKED ISSUE #315: Limit the hours per day. At the moment an employee can enter e.g. 72 hours into a single day.
TimesheetEntry validation should check for a sensical range 0-24 hours https://github.com/three-consulting/epoc-frontend/blob/main/components/form/TimesheetEntryForm.tsx#L60

The input field should have min and max properties defined https://github.com/three-consulting/epoc-frontend/blob/main/components/form/TimesheetEntryForm.tsx#L131-L138

venlavanhalathree:
closes #315

### DIFF
--- a/components/common/FormFields.tsx
+++ b/components/common/FormFields.tsx
@@ -16,7 +16,7 @@ import {
 import styled from "@emotion/styled"
 import ErrorAlert from "./ErrorAlert"
 import { AddCustomerForm } from "../form/CustomerForm"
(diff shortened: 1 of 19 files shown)


LeadingMoominExpert:
Automatic checks are failing, because of some formatting issues. Run
```bash
yarn eslint .
```
to check what's wrong and fix the issues. Then commit those changes and everything should be in order.

PR #318, opened 2023-05-23
Add first page to exported pdf reports

### LINKED ISSUE #316: Add first page to exported pdf reports
When exporting a pdf from reports-page, the exported pdf should first contain a page with all employees listed with their hours and sum of total hours of all employees

venlavanhalathree:
closes #316

### DIFF
--- a/components/table/ReportTable/utils.ts
+++ b/components/table/ReportTable/utils.ts
@@ -9,11 +9,14 @@ import {
 } from "@/lib/types/apiTypes"
 import { downloadFile } from "@/lib/utils/common"
 import { getText } from "@/lib/utils/fetch"
+import { getAllByDisplayValue } from "@testing-library/react"
 import { User } from "firebase/auth"
 import jsPDF from "jspdf"
 import autoTable from "jspdf-autotable"
 import { groupBy, round, toArray } from "lodash"
 import { DateTime } from "luxon"
+
+type EmployeeHours = [name: string, hours: number]
 
 export const customersByEmployeeTimesheets = (
     timesheets: Timesheet[],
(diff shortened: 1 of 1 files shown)


mapehe:
I could also take a look at this. Pls wait for my ✅ before merging @venlavanhalathree

components/table/ReportTable/utils.ts
```diff
@@ -152,10 +154,50 @@ export const handlePdfExportClick = (
...
     const doc = new JsPDF()
+    const startX = 14
 
     const groupedEntries = toArray(
         groupBy(timesheetEntries, (entry) => entry.timesheet.id)
     )
+
+    if (groupedEntries.length > 1) {
+        let totalHours = 0
+        const allData: EmployeeHours[] = []
```
LeadingMoominExpert, line 165: `allData` is not a very descriptive variable name, try to think of something else.

components/table/ReportTable/utils.ts
```diff
@@ -152,10 +154,50 @@ export const handlePdfExportClick = (
...
+        doc.setFontSize(20)
+        doc.setFontSize(24)
```
LeadingMoominExpert, line 167: Setting the font size twice here, first one is redundant

LeadingMoominExpert (changes requested):

components/table/ReportTable/utils.ts
```diff
@@ -152,10 +154,49 @@ export const handlePdfExportClick = (
...
+                    : 0
+            const employee = entries[0]?.timesheet?.employee
+            const employeeName = `${employee?.firstName} ${employee?.lastName}`
+            totalHours += totalHoursPerEmployee
+
+            const employeeHours: EmployeeHours = [
+                employeeName,
+                totalHoursPerEmployee,
+            ]
+            allEmployeeHours.push(employeeHours)
```
mapehe, line 188: Whatever will be left of this could be a separate function.

mapehe (changes requested):
Good job! The code works 🔥

I have some opinionated comments, which you can consider. The intention is to produce the same result, but improve the code quality.

components/table/ReportTable/utils.ts
```diff
@@ -152,10 +154,49 @@ export const handlePdfExportClick = (
...
     const doc = new JsPDF()
+    const startX = 14
 
     const groupedEntries = toArray(
         groupBy(timesheetEntries, (entry) => entry.timesheet.id)
     )
+
+    if (groupedEntries.length > 1) {
+        let totalHours = 0
+        const allEmployeeHours: EmployeeHours[] = []
```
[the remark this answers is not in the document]
mapehe: As a general rule of thumb: Whenever you write `let`, I would suggest reconsidering if it's possible to do the same thing with `const`. This will not always apply, but it's a good thing to keep in mind.

LeadingMoominExpert (approved):

mapehe (approved):
Nice 🙌

PR #330, opened 2023-06-08
create flex hook

mapehe:

PR #331, opened 2023-06-08
drop user context

mapehe:
There are two contexts `AuthContext` and `UserContext` which are basically the same. It makes no sense to have them separately.

I also made 'email' a required field in AuthContext. Makes it easier to use other hooks like the one in #330

### DIFF
--- a/components/common/Main.tsx
+++ b/components/common/Main.tsx
@@ -1,10 +1,11 @@
 import React, { ReactNode, useContext } from "react"
 import { Box, Flex } from "@chakra-ui/react"
 import LeftNav from "./LeftNav"
-import { AuthContext } from "@/lib/contexts/FirebaseAuthContext"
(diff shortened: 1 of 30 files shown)


components/common/Main.tsx
```diff
@@ -29,8 +30,16 @@ const Main = ({ children }: MainProps): JSX.Element => {
             justifyContent="center"
             minHeight="100vh"
         >
-            {isLarge && <LeftNav />}
-            <Box>{user ? children : <PleaseLogInPage isLarge={isLarge} />}</Box>
+            {isLarge && <LeftNav {...auth} />}
+            <Box>
+                {user && user.email ? (
```
mapehe, line 35: I think we can just assume that if a user managed to log in they also have an email

LeadingMoominExpert (approved):

PR #335, opened 2023-06-11
tidy hooks

mapehe:
* Simplify some code in hooks.
* Get rid of some older UI errors in the console.
* Update swr to 2.0:
     * We get rid of custom `useMatchMutate` and use officially supported features instead.
     * We fix a bug I created earlier in #332.


lib/utils/matchMutate.ts
```diff
@@ -1,23 +0,0 @@
-import { useSWRConfig } from "swr"
```
mapehe, line 1: Using SWR 2.0 we can get rid of this custom logic.

components/common/Buttons.tsx
```diff
@@ -8,6 +8,7 @@ import {
     IconButtonProps,
     StyleConfig,
 } from "@chakra-ui/react"
+import _ from "lodash"
```
LeadingMoominExpert, line 11:
```suggestion
import { omit } from "lodash"
```

components/common/Buttons.tsx
```diff
@@ -89,23 +90,14 @@ export const CustomButton = (props: ICustomButton): JSX.Element => {
...
-                props.hover
-                    ? {
-                          color: props.backgroundColor,
-                          backgroundColor: props.color,
-                      }
-                    : {}
-            }
-            {...props}
+            icon={<Icon as={props.icontype} boxSize="1.5rem" />}
+            {..._.omit(props, "icontype")}
```
LeadingMoominExpert, line 100:
```suggestion
            {...omit(props, "icontype")}
```

lib/hooks/swrInterface.ts
```diff
@@ -52,22 +57,21 @@ export const useGet = <T>(
     endpoint: string | null,
     deps: Endpoint[],
     user?: User,
-    params?: Record<string, string | number>
+    params?: Record<string, string | number | undefined>
 ): ApiGetResponse<T> => {
-    const key = getCacheKey(endpoint, deps, params)
+    const pars = _.pickBy(params, (x) => !_.isUndefined(x)) as Record<
```
LeadingMoominExpert, line 62:
```suggestion
    const pars = pickBy(params, (x) => !isUndefined(x)) as Record<
```

LeadingMoominExpert (approved):

PR #343, opened 2023-06-15
Rethink forms, tables, UI and UX

mapehe:
**TL;DR**:
* Redo forms
* Redo tables
* Redo UI and layout
* Redo UX
* Install [react-hook-form](https://www.react-hook-form.com/) and [react-table](https://tanstack.com/table/v8) to reduce the amount of code we need to maintain.

## Why

The current version of the front end is based on an early PoC that the actual app was supposed to be built on. What happened instead was that the remake was never done. Everything we currently have has grown around that old PoC, and most of it sucks.

A major update like this is really hard to do in an incremental way, so I chose to try and rewrite all the CRUD views at once excluding TimesheetEntries. I also left the report page untouched.

## Forms

The current forms are both over-engineered and sluggish: They render on every keystroke causing massive inefficiency. They are also hard to maintain because they were badly designed by @mapehe. With `react-hook-form` we get easily maintained efficient forms. Check out [one of the new forms](https://github.com/three-consulting/epoc-frontend/blob/spring-cleaning/components/form/TaskForm.tsx#L40-L72). I think they are way more legible than before.

## Tables

With [one generic table](https://github.com/three-consulting/epoc-frontend/blob/spring-cleaning/components/common/Table.tsx) and **react-table** we get searchable, sortable and selectable tables with pagination almost for free.

image

## Redo UI and layout

Since we already use Chakra, I think we should make heavier use of the library UI components.

image

One bonus is that the layouts are responsive by default. (Screenshot of the above view in Chrome mobile emulator)

image

## Redo UX

Currently we have separate pages for viewing and editing things, which causes weird extra friction when using the app. I suggest we get rid of all those and introduce a single `Organization` page instead. Well designed list views could be useful, if we need to operate on many items at once, but I'm leaving them out of scope here.

I also think we should control the loading states of rest requests in a more fine grained way so that the user knows exactly what is loading and when. (See below.) As a side effect the app feels like it's loading everything blazing fast compared to before.

image

## Misc

* Introduce [getAndMutate](https://github.com/three-consulting/epoc-frontend/blob/spring-cleaning/lib/hooks/swrInterface.ts#L43-L53) for making non-hook get requests that are compatible with the existing integration to swr.


.github/workflows/pull-request.yaml
```diff
@@ -26,7 +26,6 @@ jobs:
           yarn run type-check
           yarn prettier --check .
           yarn eslint .
-          yarn test
```
LeadingMoominExpert, line 29: 💯

mapehe:
Closing in favour of #344

